PR #12, opened 2024-03-26
Fix the modified date of downloaded files via s3 services (fix CMCS-277)

nitreb:


nitreb:
@mercator-ocean/digital-ocean-toolbox-team please have a look at this work

PR #13, opened 2024-03-26
feat: download file list with more information

renaudjester:
fix CMCS-311


copernicusmarine/python_interface/get.py
```diff
@@ -33,7 +33,7 @@ def get(
     filter: Optional[str] = None,
     regex: Optional[str] = None,
     file_list: Optional[Union[pathlib.Path, str]] = None,
-    download_file_list: bool = False,
+    download_file_list: Optional[str] = None,
```
renaudjester, line 36: Breaking change: but I don't really know how to make in python what I have done with click, should I just do some custom check in the get function?

copernicusmarine/command_line_interface/group_get.py
```diff
@@ -201,14 +201,18 @@ def cli_group_get() -> None:
 )
 @click.option(
     "--download-file-list",
-    type=bool,
-    is_flag=True,
-    default=False,
-    help="Option to only create a file files_to_download.txt containing "
+    type=str,
+    is_flag=False,
+    flag_value="files_to_download.txt",
```
nitreb, line 206: I find it a bit confusing to have an option that can be used with or without a value. Maybe create a `--download-output-filename` option that take the filename as a value, and when set, has the behavior of the ex-`--download-file-list` flag. That way, you can deprecate the old option, and don't have a breaking change in the Python API

copernicusmarine/command_line_interface/group_get.py
```diff
@@ -201,14 +201,18 @@ def cli_group_get() -> None:
...
+    default=None,
+    help="Option to only create a file containing "
     "the names of the targeted files instead of downloading them. "
     "It writes the file in the directory specified with the "
     "--output-directory option (default to current directory). "
-    "If specified, no other action will be performed.",
+    "The file name specified should end with '.txt' or '.csv' "
+    "else it will be name 'files_to_download.txt'. "
+    "If specified, no other action will be performed. "
+    "Please find more information in the README.",
```
nitreb, line 215:
```suggestion
    "The file name specified should end with '.txt' or '.csv' "
    "If specified, no other action will be performed. "
    "Please find more information in the README.",
```
I think the behavior should be if it does not end with `.txt` or `.csv`, just raise an error

nitreb:
Also, maybe some tests for csv and txt output

copernicusmarine/command_line_interface/group_get.py
renaudjester: I was happy with my flag not so flag option 😢

nitreb: did it worked? Because in the click doc, it says that if `flag_value`, then `is_flag=True`

renaudjester: https://click.palletsprojects.com/en/8.1.x/options/#optional-value yep :D

copernicusmarine/command_line_interface/group_get.py
```diff
@@ -200,15 +204,24 @@ def cli_group_get() -> None:
     "the files to download.",
 )
 @click.option(
-    "--download-file-list",
-    type=bool,
-    is_flag=True,
-    default=False,
-    help="Option to only create a file files_to_download.txt containing "
+    "--list-files",
```
nitreb, line 207: I think we should have something more explicit, like `--file-list-output-filename` or `--download-file-list-output-filename`. Maybe ask on the dedicated CMCS ticket. Because once we release it with this name, it's always tricky to change the name after

nitreb: @qgau an idea?

nitreb (approved):

PR #19, opened 2024-04-02
fix: Credentials are checked when user login

nitreb:


qgau (approved):

slim0 (approved):

tests/test_command_line_interface.py
```diff
@@ -1882,36 +1882,6 @@ def test_subset_dataset_part_option(self, tmp_path):
...
-            "--force-download",
-            "-o",
-            f"{tmp_path}",
-            "-f",
-            "result.nc",
-        ]
-
-        output = subprocess.run(command)
-        assert output.returncode == 0
-
```
nitreb, line 1914: This test fails and I don't understand what it tests so ciao

PR #20, opened 2024-04-02
fix: dowload file list not overwritten by create file list

renaudjester:
@nitreb what do you think? I don't like it much! It fixes the current problem. 

Do you have otherwise an idea how we could do deprecated options in python without them being replaced?


nitreb:
> @nitreb what do you think? I don't like it much! It fixes the current problem.
> 
> Do you have otherwise an idea how we could do deprecated options in python without them being replaced?

@renaudjester I'm not really sure. The way you did seems fine to me. We need to keep the two options separated, and raise a warning when the old one is used. I think it's the way it works in CLI. It allows to have two different usage (bool vs string)

nitreb (approved):

PR #31, opened 2024-04-10
Modulus return wrong value

uriii3:
[Link to error](https://mercator-ocean.atlassian.net/jira/software/c/projects/CMCS/issues/CMCS-295?jql=project%20%3D%20%22CMCS%22%20ORDER%20BY%20created%20DESC) CMCS-295.


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -340,6 +340,8 @@ def longitude_modulus(longitude: float) -> float:
     """
     Returns the equivalent longitude between -180 and 180
     """
+    if longitude == 180:
```
uriii3, line 343: @nitreb do you know what this function is used for and what inputs does it have?
Should I add tests?

nitreb: @uriii3 this function is used to convert a arbitrary longitude to its corresponding value in the range [-180, 180[. For example, a longitude of 15° also corresponds to 375°, 735°, etc. You can read more about how we deal with longitude in the `Note about longitude range` part of the README

nitreb: You might also want to look into the `_longitude_subset` function

nitreb: @renaudjester I think it's a bit misleading. A modulus between -180 and 360 means that it's either between -180 and 180, or between 0 and 360. 

Here an example of why you might want longitude offset between 0 and 360: if you want to request data in the pacific ocean (for example 170-190), and translate it between -180;180, then you end up with two intervals: [-180, -170] and [170, 180]. But this is not possible with xarray, thus translating it between 0 and 360.

I hope I'm clear, if not, we can have a call :)

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -115,16 +115,25 @@ def get_size_of_coordinate_subset(
...
+def _update_dataset_attributes(
+    dataset: xarray.Dataset,
+    maximum_longitude: float,
+):
     for coord_label in COORDINATES_LABEL["longitude"]:
         if coord_label in dataset.dims:
             attrs = dataset[coord_label].attrs
             if "valid_min" in attrs:
-                attrs["valid_min"] += 180
+                attrs["valid_min"] += maximum_longitude - 180
```
uriii3, line 126: I tried to handle the error from a deeper side. Now it should be able to handle the subsets, returning the desired intervals or ranges.
I'm a little bit unsure though on how to handle these `valid_min` and `valid_max` attributes from the dataset.

uriii3: @nitreb @renaudjester

nitreb: I think you can simply use `attrs["valid_min"] = coord.values.min()`

renaudjester: I think it can even be deleted tbh since it's gonna be updated anyway[ here](https://github.com/mercator-ocean/copernicus-marine-toolbox/blob/2011da1f031cb8f5624f9da3cb780dfbb974a3bd/copernicusmarine/download_functions/subset_xarray.py#L347)

nitreb:
Could you add some tests please? Especially the one example who failed before

tests/test_longitudes_with_modulus.py
```diff
@@ -132,6 +132,26 @@ def test_subset_with_longitude_over_antemeridian(self, tmp_path):
...
+        output = subprocess.run(command)
+
+        dataset = xarray.open_dataset(Path(tmp_path, filename_dataset))
+
+        longitudes = dataset.longitude.values
+
+        assert output.returncode == 0
+        assert not numpy.isnan(dataset.thetao.max().values.item())
+        assert longitudes.min() == -145
+        assert longitudes.max() == 180
```
uriii3, line 153: this was the case that the user reported, now it works fine!

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -115,16 +115,27 @@ def get_size_of_coordinate_subset(
...
 
 
-def _update_dataset_attributes(dataset: xarray.Dataset):
+def _update_dataset_attributes(
+    dataset: xarray.Dataset,
+    minimum_longitude_modulus: float,
+):
+    window = (
+        minimum_longitude_modulus + 180
+    )  # compute the degrees needed to move the dataset
```
renaudjester, line 124: Maybe I wouldn't put this comment 🤔

renaudjester (approved):

nitreb (approved):

PR #32, opened 2024-04-10
fix: change attribute when from elevation to depth

renaudjester:
fix [MDSOP-119](https://cms-change.atlassian.net/browse/MDSOP-119)

Actual behavior: 

native files: depth values go from 0 to 4000 and contains 'positive': 'down' attribute
ARCO format: elevation goes from -4000 to -0 and contains 'positive': 'up'  attribute
CMT: depth values go from 0 to 4000 but contains 'positive': 'up'  attribute

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -223,6 +223,7 @@ def _depth_subset(
             attrs = dataset["elevation"].attrs
             dataset = dataset.reindex(elevation=dataset.elevation[::-1])
             dataset["elevation"] = dataset.elevation * (-1)
+            attrs["positive"] = "down"
             dataset = dataset.rename({"elevation": "depth"})
             dataset.depth.attrs = attrs
         return dataset
--- a/tests/test_python_interface.py
+++ b/tests/test_python_interface.py
@@ -198,3 +198,9 @@ class TestPythonInterface:
             dataset_id="cmems_obs-oc_atl_bgc-plankton_nrt_l4-gapfree-multi-1km_P1D",
         )
         assert dataset.time.valid_min > "2024-01-01T00:00:00.000000000"
+
+    def test_subset_modify_attr_positive_to_down_for_depth(self):
+        dataset = open_dataset(
+            dataset_id="cmems_mod_arc_phy_anfc_6km_detided_P1D-m"
+        )
+        assert dataset.depth.attrs["positive"] == "down"


tests/test_python_interface.py
```diff
@@ -198,3 +198,9 @@ def test_open_dataset_with_retention_date(self):
             dataset_id="cmems_obs-oc_atl_bgc-plankton_nrt_l4-gapfree-multi-1km_P1D",
         )
         assert dataset.time.valid_min > "2024-01-01T00:00:00.000000000"
+
+    def test_subset_modify_attr_positive_to_down_for_depth(self):
+        dataset = open_dataset(
+            dataset_id="cmems_mod_arc_phy_anfc_6km_detided_P1D-m"
+        )
+        assert dataset.depth.attrs["positive"] == "down"
```
nitreb, line 206: You can also maybe complete the `test_conversion_between_elevation_and_depth` and `test_force_no_conversion_between_elevation_and_depth` tests, it should be just 1 line per test

tests/test_command_line_interface.py
```diff
@@ -1092,17 +1092,26 @@ def then_I_have_correct_sign_for_depth_coordinates_values(
...
             tmp_path, "positive"
         )
+        self.then_I_have_correct_attribute_value(tmp_path, "depth", "down")
 
     def test_force_no_conversion_between_elevation_and_depth(self, tmp_path):
         self.when_I_request_subset_dataset_with_zarr_service(tmp_path, False)
         self.then_I_have_correct_sign_for_depth_coordinates_values(
             tmp_path, "negative"
         )
+        # self.then_I_have_correct_attribute_value(tmp_path, "elevation", "up")
```
renaudjester, line 1114: Right now it seems that there is a bug with this specific dataset ie its attributes are not updated at all when transformed to ARCO. I am asking Guille why :D

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -223,12 +223,27 @@ def convert_elevation_to_depth(dataset: xarray.Dataset):
...
+        if "elevation" in dataset.dims:
+            attrs = dataset["elevation"].attrs
+            attrs["valid_min"] = dataset["elevation"].values.min()
+            attrs["valid_max"] = dataset["elevation"].values.max()
+            attrs["positive"] = "up"
+            attrs["standard_name"] = "elevation"
+            attrs["long_name"] = "Elevation"
+            attrs["units"] = "m"
+            dataset["elevation"].attrs = attrs
+        return dataset
```
renaudjester, line 241: Talked with Guille and he told me that some old dataset (created more than 5 months ago) have datacubes where the attributes are not updated and are still in depth (even if the dimension has been changed to elevation) so I am updating it here

PR #33, opened 2024-04-11
fix: do not trust env while being able to use proxy

renaudjester:
Fix [CMCS-331](https://mercator-ocean.atlassian.net/browse/CMCS-331)


.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Run tests
         env:
-          COPERNICUS_MARINE_SERVICE_USERNAME: ${{ secrets.COPERNICUS_MARINE_SERVICE_USERNAME }}
-          COPERNICUS_MARINE_SERVICE_PASSWORD: ${{ secrets.COPERNICUS_MARINE_SERVICE_PASSWORD }}
+          COPERNICUSMARINE_SERVICE_USERNAME: ${{ secrets.COPERNICUSMARINE_SERVICE_USERNAME }}
+          COPERNICUSMARINE_SERVICE_PASSWORD: ${{ secrets.COPERNICUSMARINE_SERVICE_PASSWORD }}
```
renaudjester, line 28: @nitreb if you agree with this change could you add those secrets to the repo so that I can run the tests? 🙏

nitreb: both secrets are added :)

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -64,9 +67,7 @@ class _ServiceShortName(str, Enum):
     MARINE_DATA_STORE_STAC_BASE_URL_STAGING + "/catalog.stac.json"
 )
 
-MAX_CONCURRENT_REQUESTS = int(
-    os.getenv("COPERNICUSMARINE_MAX_CONCURRENT_REQUESTS", "15")
-)
+MAX_CONCURRENT_REQUESTS = int(COPERNICUSMARINE_MAX_CONCURRENT_REQUESTS or "15")
```
nitreb, line 70: The default value should be where the variable is defined I think

nitreb:
There are two tickets done in this PR no? Rename and proxy stuff

.github/workflows/tests.yml
renaudjester: Super thanks!

renaudjester:
> There are two tickets done in this PR no? Rename and proxy stuff

Indeed :D

copernicusmarine/core_functions/environment_variables.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+COPERNICUSMARINE_SERVICE_USERNAME = os.getenv(
+    "COPERNICUSMARINE_SERVICE_USERNAME"
+) or os.getenv("COPERNICUS_MARINE_SERVICE_USERNAME")
```
nitreb, line 5: Should we add deprecation warnings?

nitreb (approved):

PR #36, opened 2024-04-17
Created a function to be able to compute the higher bound

uriii3:
Resolving [issue 278](https://mercator-ocean.atlassian.net/browse/CMCS-278).

To do so, I made sure that the interval was returning the correct options, adding a function to be able to compute higher bounds. With the code as is it right now, it raises warning's when the user request is out of bounds.

I would suggest, though, that if the subset is complete (from [-180, 180[) this warning is not raised. This is actually not working this way, because if the user request is [-180,180] is will never get to 180... I think the fix could be useful.

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -367,13 +367,22 @@ def subset(
 
 def longitude_modulus(longitude: float) -> float:
     """
-    Returns the equivalent longitude between -180 and 180
+    Returns the equivalent longitude between -180 and 179.999999999
     """
     # We are using Decimal to avoid issue with rounding
     modulus = float(Decimal(str(longitude + 180)) % 360)
     # Modulus with python return a negative value if the denominator is negative
     # To counteract that, we add 360 if the result is < 0
     modulus = modulus if modulus >= 0 else modulus + 360
+    return modulus - 180
+
+
+def longitude_modulus_open(longitude: float) -> float:
+    """
+    Returns the equivalent longitude between -179.9999999 and 180
+    """
+    modulus = float(Decimal(str(longitude + 180)) % 360)
+    modulus = modulus if modulus > 0 else modulus + 360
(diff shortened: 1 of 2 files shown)


tests/test_warnings_subset_bounds.py
```diff
@@ -0,0 +1,82 @@
+import subprocess
+
+
+class TestWarningsSubsetBounds:
```
uriii3, line 4: added another test to check that the problems reported where solved correctly

renaudjester (approved):

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -418,7 +428,9 @@ def check_dataset_subset_bounds(
                     else longitudes.min()
                 ),
                 user_maximum_coordinate_value=(
-                    longitude_modulus(dataset_subset.maximum_longitude)
```
nitreb, line 421: `longitude_modulus` is also called some places else in the code, no need to change it there?

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -361,6 +361,16 @@ def longitude_modulus(longitude: float) -> float:
     return modulus - 180
 
 
+def longitude_modulus_upper_bound(longitude: float) -> float:
+    """
+    Returns the equivalent longitude between -179.9999999 and 180
```
nitreb, line 366: Same as above, `in ]-180, 180]`

copernicusmarine/download_functions/subset_xarray.py
uriii3: no, actually the only problem is when considering the bounds, but the way it is handled in the other parts already makes sense to me!
Still, though, I would insist that the warning is only raised when we know some parts of the dataset are missing... doesn't make sense to add a warning when all the points are there (because the dataset doesn't fully go to 180 because the point is already repeated!

PR #37, opened 2024-04-18
feat: new option that allow direct download of files with get

renaudjester:


tests/test_get_direct_download.py
```diff
@@ -0,0 +1,155 @@
...
+            "--dataset-id",
+            "cmems_obs-ins_glo_phybgcwav_mynrt_na_irr",
+            "--direct-download-multiple",
+            DIRECT_DOWNLOAD_MULTIPLE_EXAMPLE,
+            "--force-download",
+            "--show-outputnames",
+            "--overwrite-output-data",
+        ]
+        self.output = execute_in_terminal(self.command)
+        assert (
```
slim0, line 53: We should also check that "history/BO/AR_PR_BO_58US.nc" file is present here if I'm not wrong.

slim0 (approved):

PR #41, opened 2024-04-22
feat: add release date to describe and filter datasets

renaudjester:
fix: [CMCS-282](https://mercator-ocean.atlassian.net/browse/CMCS-282)

- Add the release date to the describe
- Define an "official" dataset, version ie to copy the behavior of myocean where a dataset is shown only one tag
- By default hide double dissemination 
- Sort the catalogue of the describe so that the "official"/"default" version and part are first

Todo:
- [x] Needs to do some tests
- [x] Should we raise if the user wants a product with a release date in the future? I didn't do that, maybe open a ticket in the future if needed


copernicusmarine/command_line_interface/group_describe.py
```diff
@@ -65,6 +65,13 @@ def cli_group_describe() -> None:
     default=False,
     help="Include product keyword details in output.",
 )
+@click.option(
+    "--include-all-versions",
```
renaudjester, line 69: Any idea for the name of the option is welcome :D

slim0 (approved):

renaudjester:
> Mmm I first approved but tests are not successful... So I change my mind. xD

Anyway I cannot merge if the tests don't pass so don't worry haha

nitreb (approved):

PR #44, opened 2024-04-25
Fix cache issue by running the test that overwrite metadata cache first

nitreb:
@renaudjester is that something like this that you had in mind?

### DIFF
--- a/conda_environment_test.yaml
+++ b/conda_environment_test.yaml
@@ -7,3 +7,5 @@ dependencies:
   - tox==4.11.4
   - netcdf4==1.6.5
   - syrupy==4.6.1
+  - pip:
+    - pytest-order==1.2.1
--- a/tests/test_command_line_interface.py
+++ b/tests/test_command_line_interface.py
@@ -12,6 +12,7 @@ from json import loads
 from pathlib import Path
 from typing import List, Optional, Union
 
+import pytest
 import xarray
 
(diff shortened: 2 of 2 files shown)


conda_environment_test.yaml
```diff
@@ -7,3 +7,5 @@ dependencies:
   - tox==4.11.4
   - netcdf4==1.6.5
   - syrupy==4.6.1
+  - pip:
+    - pytest-order==1.2.1
```
renaudjester, line 11: what's the use in added it in the pip?

renaudjester (approved):

tests/test_command_line_interface.py
```diff
@@ -56,6 +57,7 @@ def get_file_size(filepath):
 
 
 class TestCommandLineInterface:
+    @pytest.mark.order(1)
```
renaudjester, line 60: I am wondering though: will it make it first only for this file? 
Because in the end the problem was that another test file was being done first

conda_environment_test.yaml
nitreb: it's not available on conda forge unfortunately

tests/test_command_line_interface.py
nitreb: it seems to have done it globally. I think pytest collects all the tests first, then runs them

PR #45, opened 2024-04-25
feat: direct download with file list

renaudjester:
After discussion with David, for a better design and user experience we decide that:
- `--direct-download-one` will be replaced by `--direct-download`
- `--direct-download-multiple` will be replaced by `--file-list`

So with the code, file list option will be used for direct downloads.
All te files from file list or direct download that are not found will be searched using a listing on the bucket


copernicusmarine/python_interface/get.py
```diff
@@ -71,8 +70,7 @@ def get(
         regex (str, optional): Apply a regular expression filter to the downloaded data.
         file_list (Union[pathlib.Path, str], optional): A path to a text file that list filenames line by line. Filenames must match the absolute paths of the files to download.
         create_file_list (str, optional): Option to only create a file containing the names of the the targeted files instead of downloading them. It writes the file in the directory specified with the --output-directory option (default to current directory). If specified, no other action will be performed.
-        direct_download_one (str, optional): Directly download a file using its path.The file has to be from the dataset specified with dataset_id.
-        direct_download_multiple (pathlib.Path, optional): Directly download multiple files using a file containing the paths. The file has to be from the dataset specified with dataset_id.
+        direct_download (str, optional): Directly download a file using its path.The file has to be from the dataset specified with dataset_id.
```
nitreb, line 73:
```suggestion
        direct_download (str, optional): Directly download a file using its path. The file has to be from the dataset specified with dataset_id. If some files are not found, the toolbox will list all the files on the remote server and try to match them.
```

nitreb (approved):

PR #46, opened 2024-04-25
Bounding box

uriii3:
Link to error [CMCS - 244](https://mercator-ocean.atlassian.net/browse/CMCS-244). 

I created a new flag `'bounding-box-method'` for more advanced users to be able to ask for a region that encompasses their requested area. It would have two options, `'inside'` (by default) or `'outside'`.

This is not yet implemented in Xarray (see third row of code [here](https://docs.xarray.dev/en/latest/user-guide/indexing.html#nearest-neighbor-lookups)) so the implementation looks up manually the limiting points (I will add the part of the code where it does this down below).

I still need to add the latitude (time too? it is also asked in the ticket) to the methodology but I thought that making the pull request would be useful to discuss the implementation.

Also I think it would be best to add some tests to it, right?

### DIFF
--- a/copernicusmarine/command_line_interface/group_subset.py
+++ b/copernicusmarine/command_line_interface/group_subset.py
@@ -20,10 +20,15 @@ from copernicusmarine.core_functions.deprecated import (
     DeprecatedClickOptionsCommand,
 )
 from copernicusmarine.core_functions.models import (
+    DEFAULT_BOUNDING_BOX_METHOD,
+    DEFAULT_BOUNDING_BOX_METHODS,
+    DEFAULT_BOUNDING_BOX_METHOD,
+    DEFAULT_BOUNDING_BOX_METHODS,
     DEFAULT_FILE_FORMAT,
     DEFAULT_FILE_FORMATS,
     DEFAULT_SUBSET_METHOD,
     DEFAULT_SUBSET_METHODS,
+    BoundingBoxMethod,
     FileFormat,
     SubsetMethod,
 )
@@ -207,6 +212,17 @@ def cli_group_subset() -> None:
     + 'with " " to ensure valid expression for format "%Y-%m-%d %H:%M:%S".',
 )
 @click.option(
+    "--bounding-box-method",
+    type=click.Choice(DEFAULT_BOUNDING_BOX_METHODS),
+    default=DEFAULT_BOUNDING_BOX_METHOD,
+    help=(
+        "The bounding box method when requesting the dataset. If inside, "
+        "(by default) it will return all points interior to the area. "
+        "If outside, it will return all the data such that the box is  "
+        "fully included. Check the documentation for more details."
+    ),
+)
+@click.option(
     "--subset-method",
     type=click.Choice(DEFAULT_SUBSET_METHODS),
     default=DEFAULT_SUBSET_METHOD,
@@ -376,6 +392,7 @@ def subset(
     vertical_dimension_as_originally_produced: bool,
     start_datetime: Optional[datetime],
     end_datetime: Optional[datetime],
+    bounding_box_method: BoundingBoxMethod,
     subset_method: SubsetMethod,
     output_filename: Optional[str],
     file_format: FileFormat,
@@ -429,6 +446,7 @@ def subset(
         vertical_dimension_as_originally_produced,
         start_datetime,
         end_datetime,
+        bounding_box_method,
         subset_method,
         output_filename,
         file_format,
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -1,6 +1,6 @@
 import logging
 import typing
-from datetime import datetime
+from datetime import datetime, timezone
 from decimal import Decimal
 from typing import List, Literal, Optional, Union
 
@@ -21,7 +21,10 @@ from copernicusmarine.core_functions.exceptions import (
     MinimumLongitudeGreaterThanMaximumLongitude,
     VariableDoesNotExistInTheDataset,
 )
-from copernicusmarine.core_functions.models import SubsetMethod
+from copernicusmarine.core_functions.models import (
+    BoundingBoxMethod,
+    SubsetMethod,
+)
 from copernicusmarine.core_functions.utils import (
     convert_datetime64_to_netcdf_timestamp,
 )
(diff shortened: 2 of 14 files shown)


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -181,6 +185,13 @@ def _longitude_subset(
                 if maximum_longitude_modulus < minimum_longitude_modulus:
                     maximum_longitude_modulus += 360
                     dataset = _update_dataset_attributes(dataset)
+                if bounding_box == "outside":
+                    minimum_longitude_modulus = dataset.sel(
+                        longitude=minimum_longitude_modulus, method="pad"
+                    ).longitude.values
+                    maximum_longitude_modulus = dataset.sel(
```
uriii3, line 192: this is actually the change, checking how the limits behave and expanding them if needed

nitreb: To be more generic and apply the rule to every dimensions, isn't it better to apply the logic in the `_dataset_custom_sel` function? Maybe `_nearest_neighbor_coordinates` could use some rework too

nitreb:
I think an example in the README could be nice, and tests too yes :)

copernicusmarine/download_functions/subset_xarray.py
uriii3: definitely yes! the `_nearest_neighbor_coordinates` function is supposed to return the only "point", right? what do you mean by reworking it?

nitreb: Yes exactly. If I remember right, it was used when a user requests a single point that does not fall into a real value. For example, --min-longitude 10 --max-longitude 10, but the real longitude is 10.1

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -73,19 +73,37 @@ def _dataset_custom_sel(
...
+                    logger.info(coord_selection.stop)
                     external_minimum = dataset.sel(
                         {coord_label: coord_selection.start}, method="pad"
                     )[coord_label].values
                     external_maximum = dataset.sel(
                         {coord_label: coord_selection.stop}, method="backfill"
                     )[coord_label].values
+                    logger.info("després de sel")
+                    logger.info(coord_label)
+                    if coord_label == "time":
```
uriii3, line 91: time added, I still need to add the tests and the examples!

renaudjester: If it concerns also time and depth does it still make sense to call the option "bounding box"? I don't know it's just that I see bounding box a lot for lon, lat stuff

uriii3: Mmh good question... I haven't added it for depth, only Lon, Lat and Time. In my head the difference makes sense but I'm definitely open to proposals!

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -65,9 +68,43 @@ def _dataset_custom_sel(
     coord_type: Literal["latitude", "longitude", "depth", "time"],
     coord_selection: Union[float, slice, datetime, None],
     method: Union[str, None] = None,
+    bounding_box: BoundingBoxMethod = "inside",
 ) -> xarray.Dataset:
     for coord_label in COORDINATES_LABEL[coord_type]:
         if coord_label in dataset.dims:
+            if bounding_box == "outside":
+                if (
+                    isinstance(coord_selection, slice)
```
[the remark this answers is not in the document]
uriii3: what do you mean by this? I'm not sure I understand your point

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -65,9 +68,43 @@ def _dataset_custom_sel(
     coord_type: Literal["latitude", "longitude", "depth", "time"],
     coord_selection: Union[float, slice, datetime, None],
     method: Union[str, None] = None,
+    bounding_box: BoundingBoxMethod = "inside",
```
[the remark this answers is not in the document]
uriii3: I wasn't sure also what was the standard about defining default values or not... does it help the other programmers, when the default it's defined? or is it just to make sure at the start that the value is not empty?

copernicusmarine/command_line_interface/group_subset.py
```diff
@@ -206,6 +209,17 @@ def cli_group_subset() -> None:
...
 )
+@click.option(
+    "--bounding-box-method",
+    type=click.Choice(DEFAULT_BOUNDING_BOX_METHODS),
+    default=DEFAULT_BOUNDING_BOX_METHOD,
+    help=(
+        "The bounding box method when requesting the dataset. If inside, "
+        "(by default) it will return the corresponding xarray selection of the limits. "
+        "If outside, it will return the limit points with the backfill or padding "
+        "(so that the returned bounding box encompasses the requested bounding box)."
```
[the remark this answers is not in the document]
uriii3: done!!

copernicusmarine/download_functions/subset_xarray.py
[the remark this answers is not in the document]
uriii3: okey I think I got it hehe 👍

tests/test_python_interface.py
```diff
@@ -206,3 +207,33 @@ def test_subset_modify_attr_for_depth(self):
...
+            "--variable",
+            "thetao",
+            "--minimum-longitude",
+            "0.01",
+            "--maximum-longitude",
+            "2.97",
+            "--minimum-latitude",
+            "0.01",
+            "--maximum-latitude",
+            "3.1",
```
uriii3, line 226: I've seen some unusual behaviour in the `time` dimension that I would like to check with @renaudjester !

copernicusmarine/download_functions/subset_xarray.py
uriii3: Okey, I've been looking at it but I'm a bit scared I'm not covering all the cases... are there tests for the problems or something like that? To know what to mimic or what to expect.

uriii3:
okey, I have solved (with @renaudjester 's help) the bug in the time dimension, I have added some tests and it looks like everything works fine. 
When asking for data that the dataset doesn't have, it breaks (as normal). I don't think that is a problem because people who play with this function should know the limits of the datasets and be a little careful, but I understand that is also my point of view so we can discuss about it.
Also the way I'm raising the error might not be the best possible, I will put it down in the conversation.
I'm definitely open to suggestions on how we could finish the implementation!

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -43,48 +46,83 @@
...
+                        )  # update the slic¡ing
+                    except Exception as e:
+                        logger.error(e)
+                        raise BoundingBoxError(
+                            "BoundingBoxMethod 'outside' "
+                            "is a method implemented "
+                            "to enlarge the selection, but it failed. "
+                            f"Make sure your selection '{coord_selection}' "
+                            "is valid when enlarged."
+                        )
```
uriii3, line 101: this is where the error I raised, I'm definitely a bit insecure about the way it works but I hoped that you could give me some feedback.

renaudjester: How much do you enlarge it? 
Shouldn't you enlarge only as much as the last value?

uriii3: yees! for example, though, in the case you ask for interval [170, 179.2] and the dataset has [-180, 179.18]. 

It will be a small amount of change (and a weird case) but the code would break when trying to get something bigger than 179.2 and not when checking the bounds in `check_dataset_subset_bounds`.

nitreb (approved):

uriii3:
I've moved everything because I messed up a bit with the GitHub merges and rebases. There is another branch open in [Bounding Box Method - Pull Request 66](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/66).

PR #48, opened 2024-04-25
Add more explicit message when catalog parsing fails

nitreb:
Relates to [CMCS-193](https://mercator-ocean.atlassian.net/browse/CMCS-193)

### DIFF
--- a/copernicusmarine/catalogue_parser/catalogue_parser.py
+++ b/copernicusmarine/catalogue_parser/catalogue_parser.py
@@ -864,7 +864,10 @@ def parse_catalogue(
         )
     except ValueError as e:
         logger.debug(f"Error while parsing catalogue: {e}")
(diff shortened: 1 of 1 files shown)


slim0 (approved):

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -864,7 +864,10 @@ def parse_catalogue(
         )
     except ValueError as e:
         logger.debug(f"Error while parsing catalogue: {e}")
-        logger.debug("Retrying parsing catalogue without cache...")
+        logger.debug(
+            "Now retrying without cache. If the problem persist, "
```
renaudjester, line 868: Maybe something like:
"If the problem with the cache persists,"
To make it clear that if there is a problem with the parse catalog (and not the cache) it won't help 
There is also a little typo 🤭

PR #59, opened 2024-05-02
Coordinates out of dataset bounds error message

uriii3:
In error [CMCS-243](https://mercator-ocean.atlassian.net/browse/CMCS-243) the user asks for a implementation in the code such that the exceptions are properly handled.

I've looked at it (the error and the proposed solution) and everything makes sense. I am just implementing the changes proposed.

I also added the same change in the way the error `VariableDoesNotExistInTheDataset`. Checking it, it had the same problem when raising the error.


uriii3:
@nitreb can you review this? I think it's a minor change.

copernicusmarine/core_functions/exceptions.py
renaudjester, line 1: Maybe add a test so you can check that you indeed get this message?

slim0 (approved):

nitreb (approved):

tests/test_python_interface.py
```diff
@@ -206,3 +207,15 @@ def test_subset_modify_attr_for_depth(self):
         assert dataset.depth.attrs["positive"] == "down"
         assert dataset.depth.attrs["standard_name"] == "depth"
         assert dataset.depth.attrs["long_name"] == "Depth"
+
+    def test_error_Coord_out_of_dataset_bounds(self):
+        try:
+            output = subset(
```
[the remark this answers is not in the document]
uriii3: why do you say so?

tests/test_python_interface.py
```diff
@@ -206,3 +207,17 @@ def test_subset_modify_attr_for_depth(self):
...
+                dataset_id="cmems_mod_glo_phy_anfc_0.083deg_P1D-m",
+                start_datetime=datetime.today() + timedelta(10),
+                force_download=True,
+                end_datetime=datetime.today()
+                + timedelta(days=10, hours=23, minutes=59),
+            )
+            assert "bb" not in "bbb"
+            assert output is not None
+        except core_functions.exceptions.CoordinatesOutOfDatasetBounds as e:
+            assert "Some or all of your subset selection" in e.__str__()
```
uriii3, line 223: does this solve it?

tests/test_raising_error_python_interface.py
```diff
@@ -0,0 +1,20 @@
...
+        ) as context:
+            subset(
+                dataset_id="cmems_mod_glo_phy_anfc_0.083deg_P1D-m",
+                start_datetime=datetime.today() + timedelta(10),
+                force_download=True,
+                end_datetime=datetime.today()
+                + timedelta(days=10, hours=23, minutes=59),
+            )
+
+        assert "Some or all of your subset selection" in str(context.exception)
```
uriii3, line 20: hope this solves the issue, it is based in [this post](https://stackoverflow.com/questions/129507/how-do-you-test-that-a-python-function-throws-an-exception)

renaudjester: I think it was better before 🤔 a bit clearer :D

renaudjester: And to solve the format issue you just needed to do: 
`_ = subset(...`
instead of
`output = subset(...`

uriii3: ugh okey, I'll change it back to where it was again then, sorry

tests/test_python_interface.py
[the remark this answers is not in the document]
renaudjester: Because it seems that what you did was adding this message:

```
f"The variable '{variable}' is neither a variable or a standard name in "
            f"the dataset."
```

But it's no what you test right?

renaudjester (approved):

PR #60, opened 2024-05-03
default _Fillvalue attribute

uriii3:
Fixing the default _FillValue attribute reported in error [CMCS-361](https://mercator-ocean.atlassian.net/browse/CMCS-361).

Inspired in the solution proposed in the same ticket and in [this stack-overflow issue](https://stackoverflow.com/questions/45693688/xarray-automatically-applying-fillvalue-to-coordinates-on-netcdf-output/45696423#45696423).

When downloading the data, before passing it to the NetCDF file, the encoding is applied. 

Hope the definition of the dictionary is good enough with the type-written format.

### DIFF
--- a/copernicusmarine/download_functions/common_download.py
+++ b/copernicusmarine/download_functions/common_download.py
@@ -58,19 +58,22 @@ def _prepare_download_dataset_as_netcdf(
 ):
     logger.debug("Writing dataset to NetCDF")
     encoding: dict[str, Any]
+    encoding = {f"{coord}": {"_FillValue": None} for coord in dataset.coords}
     if netcdf_compression_enabled:
         complevel = (
             1 if netcdf_compression_level is None else netcdf_compression_level
         )
         logger.info(f"NetCDF compression enabled with level {complevel}")
         comp = dict(
-            zlib=True, complevel=complevel, contiguous=False, shuffle=True
+            zlib=True,
+            complevel=complevel,
+            contiguous=False,
+            shuffle=True,
+            _FillValue=None,
         )
-        encoding = {var: comp for var in dataset.data_vars}
-    else:
-        encoding = {
-            f"{coord}": {"_FillValue": None} for coord in dataset.coords
-        }
+        encoding_vars = {var: comp for var in dataset.data_vars}
+        encoding.update(encoding_vars)
+
     xarray_download_format = "NETCDF3_CLASSIC" if netcdf3_compatible else None
     return dataset.to_netcdf(
         output_path,
--- a/copernicusmarine/download_functions/common_download.py
+++ b/copernicusmarine/download_functions/common_download.py
@@ -1,6 +1,6 @@
 import logging
 import pathlib
-from typing import Optional
+from typing import Any, Optional
 
 import xarray
 import zarr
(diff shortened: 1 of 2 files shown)


uriii3:
@nitreb this one too hehe! Thanks!

copernicusmarine/download_functions/common_download.py
renaudjester, line 1: Maybe some tests also otherwise lgtm

copernicusmarine/download_functions/common_download.py
```diff
@@ -57,6 +57,7 @@ def _prepare_download_dataset_as_netcdf(
     netcdf3_compatible: bool,
 ):
     logger.debug("Writing dataset to NetCDF")
+    encoding: dict[str, Any]
     if netcdf_compression_enabled:
         complevel = (
```
renaudjester, line 62: Why you don't do it if the compression level is enabled?

copernicusmarine/download_functions/common_download.py
```diff
@@ -67,7 +68,9 @@ def _prepare_download_dataset_as_netcdf(
         )
         encoding = {var: comp for var in dataset.data_vars}
```
uriii3, line 69: I thought that this line of code was creating the encoding method so that there was no need to change it. I will check that it works as expected though!

copernicusmarine/download_functions/common_download.py
uriii3: I've answered referring to another part of the code! [here](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/60/files#r1590683756)

nitreb (approved):

copernicusmarine/download_functions/common_download.py
uriii3: so, when the `netcdf_compression_enabled` was set to `True` it had the same problem! Now it is corrected! Thanks for pointing it out!

tests/test_python_interface.py
```diff
@@ -206,3 +208,25 @@ def test_subset_modify_attr_for_depth(self):
...
+            start_datetime="2024-02-23T00:00:00",
+            end_datetime="2024-02-23T23:59:59",
+            minimum_depth=0,
+            maximum_depth=1,
+            force_download=True,
+            output_filename="netcdf_fillval.nc",
+        )
+
+        subsetdata = xarray.open_dataset("netcdf_fillval.nc", decode_cf=False)
+        print(f"longitude attrs: {subsetdata.longitude.attrs}")
```
renaudjester, line 230: print

tests/test_python_interface.py
```diff
@@ -206,3 +208,25 @@ def test_subset_modify_attr_for_depth(self):
         assert dataset.depth.attrs["positive"] == "down"
         assert dataset.depth.attrs["standard_name"] == "depth"
         assert dataset.depth.attrs["long_name"] == "Depth"
+
+    def test_subset_keeps_fillvalue_empty(self):
```
renaudjester, line 212: maybe there is a place where we test the compression and you can check the value of the fillvalue also

uriii3: there is no test with the flag `netcdf_compression_enabled`, is that possible? Should I just add another tests (almost the same one) with the flag set as `True`?

renaudjester: Yes please :D

uriii3: added here: [new PR](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/62).

PR #64, opened 2024-05-08
chore: add parameters in calls to mds

renaudjester:


copernicusmarine/core_functions/utils.py
```diff
@@ -24,6 +24,7 @@
 import xarray
 from requests import PreparedRequest
 
+import copernicusmarine
```
nitreb, line 27: Isn't it better to do something like that?
```suggestion
from importlib.metadata import version
__version__ = version("copernicusmarine")
```
It avoids to import the full package

nitreb (approved):

PR #66, opened 2024-05-08
bounding Box method

uriii3:
I messed up the old branch with the bounding_box_method and I just started again, all the changes in theory have been made. We should though add the change in the naming and some minor changes also discussed with Renaud that are still pending.

PR #72, opened 2024-05-16
Adding a warning also for the depth dimension

uriii3:
Via the #66 (bounding box method) we saw that the warning for the depth dimension was not properly set.

This is an attempt to improve this feature on the toolbox.


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -492,9 +513,9 @@ def _check_coordinate_overlap(
     is_strict: bool,
 ) -> None:
     message = (
-        f"Some or all of your subset selection "
+        f"Some of your subset selection "
         f"[{user_minimum_coordinate_value}, {user_maximum_coordinate_value}] "
-        f"for the {dimension} dimension  exceed the dataset coordinates "
+        f"for the {dimension} dimension exceed the dataset coordinates "
         f"[{dataset_minimum_coordinate_value}, "
```
uriii3, line 519: I've updated the warning message, there was a typo (two spaces) and also the 'or all' didn't make sense to me, but definitely is something that can be changed back.

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -473,6 +473,27 @@ def check_dataset_subset_bounds(
...
+                else depths.max()
+            )
+            _check_coordinate_overlap(
+                dimension="depth",
+                user_minimum_coordinate_value=user_minimum_coordinate_value,
+                user_maximum_coordinate_value=user_maximum_coordinate_value,
+                dataset_minimum_coordinate_value=depths.min(),
+                dataset_maximum_coordinate_value=depths.max(),
+                is_strict=subset_method == "strict",
+            )
```
uriii3, line 496: I'm not sure that this can't be done within a function, maybe I'll try to do so. Seems very repetitive the way we are addressing the Lon-lat-depth (maybe time is more different).

uriii3:
the failed tests come from not outputting the warning message as before, so if you are okey with the update I will change all of them!

copernicusmarine/catalogue_parser/request_structure.py
```diff
@@ -34,11 +34,13 @@
 
 
 @dataclass
-class DatasetTimeAndGeographicalSubset:
+class DatasetTimeandSpaceSubset:
```
renaudjester, line 37:
```suggestion
class DatasetTimeAndSpaceSubset:
```

renaudjester: And everywhere else :D

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -473,6 +473,27 @@ def check_dataset_subset_bounds(
                 dataset_maximum_coordinate_value=dataset_maximum_coordinate_value,
                 is_strict=subset_method == "strict",
             )
+    for coordinate_label in COORDINATES_LABEL["depth"]:
```
renaudjester, line 476: Maybe add some tests about this 🤭

tests/test_warnings_subset_bounds.py
```diff
@@ -114,15 +114,62 @@ def test_subset_strict_error(self):
...
+        output_filename = "output.nc"
+        min_longitude = 29.0
+        max_longitude = 30.0
+        min_latitude = 30
+        max_latitude = 32
+        min_depth = 0.4
+        max_depth = 50.0
+        start_datetime = "2021-11-03"
+        end_datetime = "2021-11-03"
+        command = [
```
renaudjester, line 140: you can use the "build_custom_command" function instead like above and also you don't need to download so the `--force-download` is not necessary :D

copernicusmarine/download_functions/subset_xarray.py
uriii3: it wasn't pretty

renaudjester:
> Via the #66 (bounding box method) we saw that the warning for the depth dimension was not properly set.
> 
> This is an attempt to improve this feature on the toolbox.

on hold for v2

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -510,6 +510,25 @@ def check_dataset_subset_bounds(
...
+        if coordinate_label in dataset.sizes:
+            depths = -1 * dataset_coordinates[coordinate_label].values
+            _check_coordinate_overlap(
+                dimension="depth",
+                user_minimum_coordinate_value=(
+                    dataset_subset.minimum_depth
```
renaudjester, line 519: you need to check what the user asked for. Because if the user asked for `originally-produced-data` then I think the input interval is in elevation and not in depths.

tests/test_longitudes_with_modulus.py
```diff
@@ -19,9 +19,9 @@ def _build_custom_command(
             "--variable",
             "thetao",
             "--minimum-depth",
-            "0",
+            "0.5",
```
renaudjester, line 22: why that?

tests/test_warnings_subset_bounds.py
renaudjester, line 1: Maybe you can get all the tests that tests. the warning for subset bounds and put them here so we centralise and don't repeat

tests/test_warnings_subset_bounds.py
```diff
@@ -126,3 +126,45 @@ def test_subset_handle_180_point_correctly(self):
         assert (
             b"Some or all of your subset selection"
         ) not in self.output.stderr
+
+    def test_warn_depth_out_of_dataset_bounds(self, tmp_path):
```
renaudjester, line 130: maybe add a test when you requested the elevation

copernicusmarine/download_functions/subset_xarray.py
uriii3: I've tried `--vertical-dimension-as-originally-produced True` in a `copernicusmarine` call and it seems as it is working as expected. I'm going to introduce now another test to have it clear.

tests/test_longitudes_with_modulus.py
renaudjester: I don't understand what it breaks?

tests/test_warnings_subset_bounds.py
uriii3: Okey, I've put them altogether in the same file hehe.

tests/test_longitudes_with_modulus.py
uriii3: The call. It returns as an error (you are asking for something outside of bounds.

tests/test_warnings_subset_bounds.py
uriii3: Done!

renaudjester (approved):

PR #75, opened 2024-05-17
fix: fillvalue for depth was nan

renaudjester:
depth fill_value was nan


slim0 (approved):

uriii3:
So, for `depth` the attribute `_FillValue` might not exist at the start and the default option would be `nan`?

uriii3 (approved):

PR #77, opened 2024-05-21
fix: change astype int to use datetime

renaudjester:
Seems like `astype(int)` from `numpy.datetime64` is returning completely different results on Windows than on Unix creating two bugs
https://cms-change.atlassian.net/browse/CMT-9
and https://cms-change.atlassian.net/browse/CMT-8

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -65,11 +65,6 @@ def _dataset_custom_sel(
                 nearest_neighbor_value = dataset.sel(
                     {coord_label: target}, method="nearest"
                 )[coord_label].values
-                if coord_label == "time":
-                    nanosecond = 1e-9
-                    nearest_neighbor_value = datetime.fromtimestamp(
-                        nearest_neighbor_value.astype(datetime) * nanosecond
-                    )
                 dataset = dataset.sel(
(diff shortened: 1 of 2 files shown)


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -65,11 +65,6 @@ def _dataset_custom_sel(
                 nearest_neighbor_value = dataset.sel(
                     {coord_label: target}, method="nearest"
                 )[coord_label].values
-                if coord_label == "time":
-                    nanosecond = 1e-9
-                    nearest_neighbor_value = datetime.fromtimestamp(
-                        nearest_neighbor_value.astype(int) * nanosecond
-                    )
```
uriii3, line 72: If I remember correctly, this part was useful because when doing `dataset.sel(...)`, the returned value was not "compliant" to be in the slice to select again... with the change in the function in `utils.py` this has been solved?

uriii3 (approved):

copernicusmarine/download_functions/subset_xarray.py
uriii3: okey, seems like I was seeing ghosts... it works perfectly!

nitreb (approved):

PR #84, opened 2024-06-13
fix: send logs to stderr and logging conf as a py file

renaudjester:
Fixes: https://cms-change.atlassian.net/browse/CMT-10
And: https://cms-change.atlassian.net/browse/CMT-56


tests/test_command_line_interface.py
```diff
@@ -253,17 +257,23 @@ def then_I_can_read_the_filtered_json(self):
 
     def when_I_run_copernicus_marine_describe_including_datasets(self):
         command = "copernicusmarine describe --include-datasets"
-        self.output = subprocess.check_output(command, shell=True)
+        command = [
```
uriii3, line 260: do we need to define 2 commands? or is it a typo?

tests/test_command_line_interface.py
```diff
@@ -1847,15 +1842,17 @@ def test_static_arco_service(self, tmp_path):
...
         ]
 
-        output = subprocess.run(base_command, capture_output=True)
-        assert output.returncode == 0
-        assert b"Downloading using service static-arco..." in output.stdout
+        self.output = execute_in_terminal(base_command)
+        assert self.output.returncode == 0
+        assert (
+            b"Downloading using service static-arco..." in self.output.stderr
+        )
```
uriii3, line 1849: this is info, no? is it going too to `stderr`?

renaudjester: Because it seems like all logs should go to stderr: see https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout

tests/test_command_line_interface.py
renaudjester: typo :D

PR #86, opened 2024-06-27
test: adapting the test to the new release

uriii3:
With the latest release of the Marine Data Store, the tests needed to be updated. Maybe we could find a way to better standardise the procedure.


tests/test_get_index_files_insitu.py
```diff
@@ -11,12 +11,12 @@ def test_get_index_insitu_files(self):
...
             "--index-parts",
         ]
-        self.output = execute_in_terminal(self.command)
+        self.output = execute_in_terminal(self.command, input=b"n\n")
 
         assert (
             b"s3://mdl-native-01/native/INSITU_GLO_PHYBGCWAV_DISCRETE_MYNRT_013_030"
             b"/cmems_obs-ins_glo_phybgcwav_mynrt_na_irr_202311/index_reference.txt"
-            in self.output.stdout
+            not in self.output.stdout
```
renaudjester, line 19: This we should either get rid of the test or find another example, because with the "not in" we are not testing anything

uriii3: Yep, cool, I see your point!

PR #93, opened 2024-07-10
chore: updating the attributes for the variables, coordinates and dataset

uriii3:
Trying to merge into the version 2 of the toolbox. Adapting the output file to be compatible with the netcdf standards.

### DIFF
--- a/copernicusmarine/catalogue_parser/catalogue_parser.py
+++ b/copernicusmarine/catalogue_parser/catalogue_parser.py
@@ -179,8 +179,11 @@ class CopernicusMarineCoordinates:
 class CopernicusMarineVariable:
     short_name: str
     standard_name: str
+    # long_name: str
     units: str
     bbox: Tuple[float, float, float, float]
+    # validMin: Optional[float]
+    # validMax: Optional[float]
     coordinates: list[CopernicusMarineCoordinates]
 
 
@@ -628,8 +631,11 @@ def _get_variables(
         CopernicusMarineVariable(
             short_name=var_cube["id"],
             standard_name=var_cube["standardName"],
+            # long_name=var_cube["name"]["en"],
             units=var_cube.get("unit") or "",
             bbox=bbox,
+            # validMin=var_cube.get("valueMin"),
+            # validMax=var_cube.get("valueMax"),
             coordinates=_get_coordinates(
                 var_cube["id"],
                 stac_asset,
--- a/copernicusmarine/catalogue_parser/catalogue_parser.py
+++ b/copernicusmarine/catalogue_parser/catalogue_parser.py
@@ -179,11 +179,8 @@ class CopernicusMarineCoordinates:
 class CopernicusMarineVariable:
     short_name: str
     standard_name: str
-    # long_name: str
     units: str
     bbox: Tuple[float, float, float, float]
-    # validMin: Optional[float]
-    # validMax: Optional[float]
     coordinates: list[CopernicusMarineCoordinates]
 
 
@@ -631,11 +628,8 @@ def _get_variables(
         CopernicusMarineVariable(
             short_name=var_cube["id"],
             standard_name=var_cube["standardName"],
-            # long_name=var_cube["name"]["en"],
             units=var_cube.get("unit") or "",
             bbox=bbox,
-            # validMin=var_cube.get("valueMin"),
-            # validMax=var_cube.get("valueMax"),
             coordinates=_get_coordinates(
                 var_cube["id"],
                 stac_asset,
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -314,14 +314,9 @@ def _update_variables_attributes(
     dataset: xarray.Dataset, variables: List[str]
 ) -> xarray.Dataset:
     for variable in variables:
-        attributes_to_keep = list(
-            set(NETCDF_CONVENTION_VARIABLE_ATTRIBUTES).intersection(
-                dataset[variable].attrs
-            )
+        dataset[variable].attrs = _filter_attributes(
+            dataset[variable].attrs, NETCDF_CONVENTION_VARIABLE_ATTRIBUTES
         )
-        dataset[variable].attrs = {
-            key: dataset[variable].attrs[key] for key in attributes_to_keep
-        }
     return dataset
 
 
@@ -347,6 +342,11 @@ def _variables_subset(
     return _update_variables_attributes(dataset, dataset_variables_filter)
 
 
+def _filter_attributes(attributes: dict, attributes_to_keep: List[str]):
+    attributes_that_exist = set(attributes).intersection(attributes_to_keep)
+    return {key: attributes[key] for key in attributes_that_exist}
+
+
 def _update_dataset_coordinate_attributes(
     dataset: xarray.Dataset,
 ) -> xarray.Dataset:
@@ -355,7 +355,9 @@ def _update_dataset_coordinate_attributes(
             if coordinate_alias in dataset.sizes:
                 coord = dataset[coordinate_alias]
                 attrs = coord.attrs
-                coordinate_attributes = NETCDF_CONVENTION_COORDINATE_ATTRIBUTES
+                coordinate_attributes = (
+                    NETCDF_CONVENTION_COORDINATE_ATTRIBUTES.copy()
+                )
                 if "time" in coordinate_label:
                     min_time_dimension = coord.values.min()
                     max_time_dimension = coord.values.max()
@@ -381,20 +383,12 @@ def _update_dataset_coordinate_attributes(
                 elif coordinate_label == "longitude":
                     coordinate_attributes.remove("valid_min")
                     coordinate_attributes.remove("valid_max")
-                attributes_to_keep = set(coordinate_attributes).intersection(
-                    attrs
-                )
-                coord.attrs = {key: attrs[key] for key in attributes_to_keep}
-    return _update_dataset_attrs(dataset)
+                coord.attrs = _filter_attributes(attrs, coordinate_attributes)
 
+    dataset.attrs = _filter_attributes(
+        dataset.attrs, NETCDF_CONVENTION_DATASET_ATTRIBUTES
+    )
 
-def _update_dataset_attrs(
-    dataset: xarray.Dataset,
-) -> xarray.Dataset:
-    attributes_to_keep = set(
-        NETCDF_CONVENTION_DATASET_ATTRIBUTES
-    ).intersection(dataset.attrs)
-    dataset.attrs = {key: dataset.attrs[key] for key in attributes_to_keep}
     return dataset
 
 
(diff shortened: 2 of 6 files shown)


copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -179,8 +179,11 @@ def convert_elevation_to_depth(self):
 class CopernicusMarineVariable:
     short_name: str
     standard_name: str
+    # long_name: str
     units: str
     bbox: Tuple[float, float, float, float]
+    # validMin: Optional[float]
+    # validMax: Optional[float]
```
renaudjester, line 186: comments

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -628,8 +631,11 @@ def _get_variables(
         CopernicusMarineVariable(
             short_name=var_cube["id"],
             standard_name=var_cube["standardName"],
+            # long_name=var_cube["name"]["en"],
             units=var_cube.get("unit") or "",
             bbox=bbox,
+            # validMin=var_cube.get("valueMin"),
+            # validMax=var_cube.get("valueMax"),
             coordinates=_get_coordinates(
```
renaudjester, line 639: here too

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -280,6 +280,28 @@ def _get_variable_name_from_standard_name(
...
+    dataset: xarray.Dataset, variables: Optional[List[str]]
+) -> xarray.Dataset:
+    list_var_attrs = [
+        "standard_name",
+        "long_name",
+        "units",
+        "unit_long",
+        "valid_min",
+        "valid_max",
+    ]
```
renaudjester, line 293: Maybe create a variable at the file level or even in the models.py file
Something like `NETCDF_CONVENTION_VARIABLE_ATTRIBUTES` or anything that makes it obvious (we don't want this to be hidden in a random function)
I am thinking maybe to keep the NETCDF_CONVENTION prefix for choices that are related to this

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -298,17 +320,32 @@ def _variables_subset(
...
             else:
                 raise VariableDoesNotExistInTheDataset(variable)
-    return dataset[numpy.array(dataset_variables_filter)]
+    dataset = dataset[
+        numpy.array(dataset_variables_filter)
+    ]  # filter the right variable
+    logger.info(f"Variables selected: {dataset_variables_filter}")
+    return _filter_variable_attrs(
+        dataset, dataset_variables_filter
+    )  # and chose the right attributes
```
renaudjester, line 329: comments

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -280,6 +280,28 @@ def _get_variable_name_from_standard_name(
     return None
 
 
+def _filter_variable_attrs(
+    dataset: xarray.Dataset, variables: Optional[List[str]]
```
renaudjester, line 284: Can variables ever be None?

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -280,6 +280,28 @@ def _get_variable_name_from_standard_name(
...
+) -> xarray.Dataset:
+    list_var_attrs = [
+        "standard_name",
+        "long_name",
+        "units",
+        "unit_long",
+        "valid_min",
+        "valid_max",
+    ]
+    if variables:
```
renaudjester, line 294: you don't need the `if variables`

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -298,17 +320,32 @@ def _variables_subset(
...
                 attrs = coord.attrs
+                list_coord_attrs = [
+                    "standard_name",
+                    "long_name",
+                    "units",
+                    "unit_long",
+                    "axis",
+                    "valid_min",
+                    "valid_max",
+                ]
```
renaudjester, line 348: Same as before with making it more visible

renaudjester: Also in general let's try to use full words for variables and here the `list` is not really needed so something like:
`coordinates_attributes` would be better

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -319,14 +356,57 @@ def _update_dataset_coordinate_valid_minmax_attributes(
...
                     attrs["valid_min"] = valid_min
                     attrs["valid_max"] = valid_max
-                else:
+                    # attrs["calendar"] = coord.encoding["calendar"]
+                    # attrs["units"] = coord.encoding["units"]
+                    attrs["axis"] = "T"
+                    attrs["unit_long"] = (
+                        coord.encoding["units"].replace("_", " ").title()
+                    )
+                    # list_coord_attrs.append("calendar")
```
renaudjester, line 369: comments

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -280,6 +280,28 @@ def _get_variable_name_from_standard_name(
...
+        "units",
+        "unit_long",
+        "valid_min",
+        "valid_max",
+    ]
+    if variables:
+        for variable in variables:
+            existing_keys = list(
+                set(list_var_attrs).intersection(dataset[variable].attrs)
+            )
```
renaudjester, line 298: you don't need to do `list`

renaudjester: Also I would name this variable something like: `attributes_to_keep` seems clearer to me

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -280,6 +280,28 @@ def _get_variable_name_from_standard_name(
     return None
 
 
+def _filter_variable_attrs(
```
renaudjester, line 283: `_update_variables_attributes` maybe better

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -298,17 +320,32 @@ def _variables_subset(
                 )
             else:
                 raise VariableDoesNotExistInTheDataset(variable)
-    return dataset[numpy.array(dataset_variables_filter)]
+    dataset = dataset[
+        numpy.array(dataset_variables_filter)
+    ]  # filter the right variable
+    logger.info(f"Variables selected: {dataset_variables_filter}")
```
renaudjester, line 326: Not sure this log is necessary?

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -319,14 +356,57 @@ def _update_dataset_coordinate_valid_minmax_attributes(
...
+                elif coordinate_label in ["latitude", "depth", "elevation"]:
                     attrs["valid_min"] = coord.values.min()
                     attrs["valid_max"] = coord.values.max()
+                    existing_keys = list(
+                        set(list_coord_attrs).intersection(attrs)
+                    )
+                    coord.attrs = {key: attrs[key] for key in existing_keys}
+                else:  # for longitude
+                    list_coord_attrs.remove("valid_min")
+                    list_coord_attrs.remove("valid_max")
```
renaudjester, line 384: They shouldn't be set right? 🤔

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -319,14 +356,57 @@ def _update_dataset_coordinate_valid_minmax_attributes(
...
+                    # attrs["units"] = coord.encoding["units"]
+                    attrs["axis"] = "T"
+                    attrs["unit_long"] = (
+                        coord.encoding["units"].replace("_", " ").title()
+                    )
+                    # list_coord_attrs.append("calendar")
+                    list_coord_attrs.remove("units")
+                    existing_keys = list(
+                        set(list_coord_attrs).intersection(attrs)
+                    )
```
renaudjester, line 373: you don't need to `list()`

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -319,14 +356,57 @@ def _update_dataset_coordinate_valid_minmax_attributes(
...
+                    coord.attrs = {key: attrs[key] for key in existing_keys}
```
renaudjester, line 374: since you do this 3 times maybe you can do it after the "if"

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -319,14 +356,57 @@ def _update_dataset_coordinate_valid_minmax_attributes(
...
+                elif coordinate_label in ["latitude", "depth", "elevation"]:
                     attrs["valid_min"] = coord.values.min()
                     attrs["valid_max"] = coord.values.max()
+                    existing_keys = list(
+                        set(list_coord_attrs).intersection(attrs)
+                    )
+                    coord.attrs = {key: attrs[key] for key in existing_keys}
+                else:  # for longitude
```
renaudjester, line 382: then just do: `elif coordinate_label == "longitude"` first

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -351,7 +431,9 @@ def subset(
 
     dataset = _depth_subset(dataset, depth_parameters)
 
-    dataset = _update_dataset_coordinate_valid_minmax_attributes(dataset)
+    dataset = _update_dataset_coordinate_attributes(dataset)
+
+    dataset = _update_dataset_attrs(dataset)
```
renaudjester, line 436: I feel like it should be in the same function

tests/test_python_interface.py
```diff
@@ -236,7 +235,6 @@ def test_subset_keeps_fillvalue_empty(self, tmp_path):
         assert "_FillValue" not in subsetdata.time.attrs
         assert "_FillValue" not in subsetdata.latitude.attrs
         assert "_FillValue" not in subsetdata.depth.attrs
-        assert "valid_max" in subsetdata.longitude.attrs
```
renaudjester, line 239: you could actually check the opposite: `assert "valid_max" not in subsetdata.longitude.attrs`

copernicusmarine/download_functions/subset_xarray.py
uriii3: mmmh if I do it before we join PR#95, there is no models.py file. I can add it manually but maybe it will then be harder and not that functional for you to merge? Or I can put them as variables at the start of the code and then when we merge we join everything?
Maybe the easiest is top create it manually and so we don't have to think about it a lot

copernicusmarine/download_functions/subset_xarray.py
uriii3: correctly, for longitude the valid_min and valid_max shouldn't be set, accordingly to [FETPRO-213](https://cms-change.atlassian.net/browse/FETPRO-213).

copernicusmarine/download_functions/subset_xarray.py
renaudjester: There is a models.py in the core_functions folders already! maybe you can create one in the download_functions folder also! I don't think this belongs to the catalog parser so it should be fine the merging

renaudjester (approved):

PR #95, opened 2024-07-11
chore: refactor to stop using the cache

renaudjester:
Fix: [CMT-2](https://cms-change.atlassian.net/browse/CMT-2) and stop using the cache

- [ ] Dataset - product mapping is ready


copernicusmarine/core_functions/subset.py
```diff
@@ -99,17 +85,17 @@ def subset_function(
...
-        subset_request = subset_request_from_file(request_file)
+        subset_request.from_file(request_file)
+    if not subset_request.dataset_id:
+        raise ValueError("Please provide a dataset id for a subset request.")
     if motu_api_request:
         motu_api_subset_request = convert_motu_api_request_to_structure(
             motu_api_request
         )
         subset_request.update(motu_api_subset_request.__dict__)
     request_update_dict = {
```
uriii3, line 98: are we sure that you can delete the update of the id? nothing weird with python variables?

copernicusmarine/core_functions/subset.py
```diff
@@ -153,30 +138,10 @@ def subset_function(
...
-                )
-            else:
-                syntax_error = SyntaxError(
-                    "Must specify at least one of "
-                    "'dataset_url' or 'dataset_id' options"
-                )
-                raise syntax_error
-        else:
-            dataset_id = subset_request.dataset_id
+            raise SyntaxError("Must specify 'dataset_id' option")
```
renaudjester, line 141: To delete

copernicusmarine/core_functions/subset.py
renaudjester: I don't think so but better be on the side of caution

renaudjester: I checked and it seems to work with the obvious cases

uriii3 (approved):

PR #96, opened 2024-07-16
fix: trying to fix the warning on the lonngitude dimension

uriii3:
Trying to fix ticket [CMT-54](https://cms-change.atlassian.net/browse/CMT-54).

Looks like a conversation about whether or not if the requested dataset is [-180, 180] and result is [-180, 180[ should still return a warning... a part from that, I think it works fine, still to be evaluated with some examples and tests.


uriii3:
I've changed some of the handling of the warnings and errors, seems clearer but also might make it a breaking change. I'll expose the procedure and the methodology in a ticket in Jira.

uriii3:
Seems like I was complicating my life, let's close it and maybe reopen if it is needed.

PR #98, opened 2024-07-17
fix: if not permission modifie last updated date on file continue

renaudjester:
fix https://cms-change.atlassian.net/browse/CMT-42

### DIFF
--- a/copernicusmarine/download_functions/download_original_files.py
+++ b/copernicusmarine/download_functions/download_original_files.py
@@ -536,7 +536,7 @@ def _download_files(
         endpoint_url: str, bucket: str, file_in: str, file_out: pathlib.Path
     ) -> pathlib.Path:
         """
-        Download ONE file and return a string of the result
+        Download ONE file and return the path of the result
         """
         s3_client, s3_resource = get_configured_boto3_session(
             endpoint_url,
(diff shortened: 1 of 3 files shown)


copernicusmarine/download_functions/download_original_files.py
```diff
@@ -554,9 +554,15 @@ def _original_files_file_download(
             file_out,
         )
 
-        os.utime(
-            file_out, (last_modified_date_epoch, last_modified_date_epoch)
-        )
+        try:
+            os.utime(
+                file_out, (last_modified_date_epoch, last_modified_date_epoch)
+            )
```
uriii3, line 560: why do we actually need to change the time of the file?

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -554,9 +554,15 @@ def _original_files_file_download(
...
+        except PermissionError:
+            logger.warning(
+                f"Permission to modify the last modified date "
+                f"of the file {file_out} is denied."
+            )
```
uriii3, line 565: as you say in the ticket, maybe a lot of warnings, but still I think it will be manageable, so doesn't bother me a lot. What do you think?

copernicusmarine/download_functions/download_original_files.py
renaudjester: Because some users want to have the modification time from the server and not from when they downloaded the file

copernicusmarine/download_functions/download_original_files.py
renaudjester: I thought that it would be okay as only a few users will have this problem

uriii3 (approved):

PR #99, opened 2024-07-25
fix: consider 180 and -180 as the same point

renaudjester:
fix: [CMT-54](https://cms-change.atlassian.net/browse/CMT-54)

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -98,7 +98,7 @@ def get_size_of_coordinate_subset(
         )
 
 
-def _update_dataset_attributes(
+def _shift_longitude_dimension(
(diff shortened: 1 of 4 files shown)


uriii3:
I'm trying to imagine what would happen if the requested interval was [-150, 190]... (it would raise a warning but still return the whole requested interval, no?)

In my mind the solution would also have to take into account this sort of requests, but maybe it doesn't have that much sense because it is not that common to do these requests... I don't know, I let you decide what you prefer!

renaudjester:
> I'm trying to imagine what would happen if the requested interval was [-150, 190]... (it would raise a warning but still return the whole requested interval, no?)
> 
> In my mind the solution would also have to take into account this sort of requests, but maybe it doesn't have that much sense because it is not that common to do these requests... I don't know, I let you decide what you prefer!

PR #100, opened 2024-07-26
fix: windows doesnt accept timestamp before 1970

renaudjester:
fix [CMT-85]( https://cms-change.atlassian.net/browse/CMT-85)

Interestingly it seems like Windows does not support date before 1970. Hence the bug described in CMT-85 where the open dataset tries to set a valid min in 1950

### DIFF
--- a/copernicusmarine/core_functions/utils.py
+++ b/copernicusmarine/core_functions/utils.py
@@ -3,7 +3,7 @@ import logging
 import os
 import pathlib
 import re
-from datetime import datetime, timezone
+from datetime import datetime
 from importlib.metadata import version
 from typing import (
     Any,
@@ -21,6 +21,7 @@ from typing import (
 
 import cftime
 import numpy
+import pandas as pd
 import xarray
 from requests import PreparedRequest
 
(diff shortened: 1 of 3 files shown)


copernicusmarine/core_functions/utils.py
```diff
@@ -148,11 +149,8 @@ def convert_datetime64_to_netcdf_timestamp(
     datetime_value: numpy.datetime64,
     cftime_unit: str,
 ) -> int:
-    nanosecond = 1e-9
-    date = datetime.fromtimestamp(
-        datetime_value.astype(datetime) * nanosecond, tz=timezone.utc
-    )
-    return cftime.date2num(date, cftime_unit)
+    pandas_datetime = pd.to_datetime(datetime_value)
+    return cftime.date2num(pandas_datetime, cftime_unit)
```
uriii3, line 153: so basically we handle the date time with pandas, no? I don't know if we should revise more parts of the code maybe where we can have this problem? on ticket [CMT-64](https://cms-change.atlassian.net/browse/CMT-64) might be a good opportunity to do so!

uriii3 (approved):

copernicusmarine/core_functions/utils.py
renaudjester: Yeah we might need to check this in CMT-64 and be sure that we have the same convention across the repo! let's keep it in mind

PR #103, opened 2024-07-30
Reading 'positive' attribute back to the depth dimension

uriii3:
After a discussion with Tony in [CMT-35](https://cms-change.atlassian.net/browse/CMT-35) we discussed that it was better to keep the attribute `positive` for the depth dimension, set correctly to `down` if the dimension was depth (by default) or to `up` if the dimension was elevation.

I considered that it was best to keep these changes separately from other ones concerning the netcdf compliance of the output files of the toolbox. This way, we can more easily undo if there is some problem or found exactly which could be the case.


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -377,7 +377,11 @@ def _update_dataset_coordinate_attributes(
                         coord.encoding["units"].replace("_", " ").title()
                     )
                     coordinate_attributes.remove("units")
-                elif coordinate_label in ["latitude", "depth", "elevation"]:
+                elif coordinate_label in ["depth", "elevation"]:
+                    attrs["valid_min"] = coord.values.min()
+                    attrs["valid_max"] = coord.values.max()
+                    coordinate_attributes.append("positive")
+                elif coordinate_label == "latitude":
```
uriii3, line 384: this might seem weird but makes quite sense and it will all be deleted with the other PR... but I wanted to keep it separated just in case.

renaudjester (approved):

PR #118, opened 2024-08-22
chore: add html documentation page to the toolbox with sphinx

renaudjester:
Fix: [CMT-17](https://cms-change.atlassian.net/browse/CMT-17)

By adding documentation. 

Missing: 

- [ ] Some action to deploy the html (github pages, read the docs, some domain?)
- [ ] Trigger the creation of the doc (precommit doesn't seem easy, otherwise maybe github action that commits)


doc/_build/html/_static/doctools.js
```diff
@@ -0,0 +1,156 @@
...
+ * doctools.js
+ * ~~~~~~~~~~~
+ *
+ * Base JavaScript utilities for all Sphinx HTML documentation.
+ *
+ * :copyright: Copyright 2007-2024 by the Sphinx team, see AUTHORS.
+ * :license: BSD, see LICENSE for details.
+ *
+ */
+"use strict";
```
uriii3, line 11: wow I have no idea about javascript hahah, but seems semi understandable the logic

doc/_static/under_construction.gif
uriii3, line 1: how come we have the same gif repeated twice on the code? is there one that is redundant?

uriii3:
Nice work, with the creation of it and the rewording to make it readable and everything, it will be awesome!

Guess we will have to discuss what you point as to do and that will be it!

PR #122, opened 2024-08-27
fix: minimum datetime should be in iso format

renaudjester:


uriii3:
is this for [CMT-100](https://cms-change.atlassian.net/browse/CMT-100) ?

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -616,7 +616,7 @@ def _format_arco_data_metadata_producer_valid_start_date(
             ).timestamp()
             * 1000
         )
-    return arco_data_metadata_producer_valid_start_date.split(".")[0]
+    return arco_data_metadata_producer_valid_start_date
```
uriii3, line 619: okay, so basically when we are not returning it to a timestamp, it is okay to maintain (and they prefer it this way) the milliseconds and everything, no?
When are we sending it to timestamp and when not? In the code it appears `isinstance(coordinates_info.get("min"), int)` this but not sure what is the logic behind.

uriii3: I mean I was checking the ones that were supposed to be timestamps and the ones that didn't and all looked the same... what is the difference behind them?

uriii3 (approved):

PR #126, opened 2024-09-04
Approximation downloaded data

renaudjester:
Fix [CMT-63](https://cms-change.atlassian.net/browse/CMT-63)


copernicusmarine/download_functions/utils.py
```diff
@@ -170,14 +176,138 @@ def _format_datetimes(
...
         coordinates_size
         * len(list(dataset.data_vars))
         * dataset[list(dataset.data_vars)[0]].dtype.itemsize
         / 1048e3
     )
-    return f"{estimate_size:.3f} MB"
+    estimated_size_message += f"{estimate_size:.3f} MB"
+
+    download_estimated_size = 0
+    for variable_name in dataset.data_vars:
```
uriii3, line 198: why do we need to iterate through the variables? do they have different points or different types of description within the same dataset? (I would expect that the see temperature and the wind might be have values over the same points, so we would only need to multiply for the number of variables, no?)

uriii3: they have different steps?

copernicusmarine/download_functions/utils.py
```diff
@@ -170,14 +176,138 @@ def _format_datetimes(
...
+        coordinates_size = 1
+        for coordinate_name in dataset.sizes:
```
uriii3, line 200: coordinate=dimension, right?

copernicusmarine/download_functions/utils.py
```diff
@@ -170,14 +176,138 @@ def _format_datetimes(
...
+            coordinate = [
+                coord
+                for coord in [
+                    var
+                    for var in service.variables
+                    if var.short_name == variable_name
+                ][0].coordinates
+                if coord.coordinate_id == coordinate_name
```
uriii3, line 208: why do we need to have a for for all the dimensions and then do them one by one? is there any case in which we are using 'lon' and then 'longitude' or something like that? or what is this logic for?

copernicusmarine/download_functions/utils.py
renaudjester: Because they might have different types (not sure you are right) 
But imaging the wind is in float64 and the thetao is in float32

copernicusmarine/download_functions/utils.py
renaudjester: Yep! but we call it coordinate in the describe

copernicusmarine/download_functions/utils.py
renaudjester: it's just that we are in a loop over the coordinates of the dataset object and what we want for this given coordinate is its metadata, so this loop is to get the metadata of the specific coordinate

renaudjester: Maybe you are right though, there could be some performance improvements haha

copernicusmarine/download_functions/utils.py
uriii3: And what would be the problem there? the amount of information per step (per chunk) would be different, but not the way to check for the amount of points to account, no? I think I'm not yet fully understanding the methodology followed, still checking the code hehe

renaudjester: It wouldn't be a problem but it would change the result.

Also some variables don't have depth, so they are lighter than others

uriii3 (approved):

PR #128, opened 2024-09-09
fix: Full tests on all datasets

renaudjester:
Fix [CMT-1](https://cms-change.atlassian.net/browse/CMT-1)


renaudjester:
Needs to be redone due to branch workflow update on the repo

PR #129, opened 2024-09-09
chore: add more return metadata and dry run

renaudjester:
Fix: [CMT-97](https://cms-change.atlassian.net/browse/CMT-97)

And adds some documentation

- [x] Add some tests?


copernicusmarine/core_functions/models.py
```diff
@@ -15,3 +18,73 @@
...
 DEFAULT_BOUNDING_BOX_METHODS = list(get_args(BoundingBoxMethod))
+
+
+class FileGet(BaseModel):
+    #: Full url of the location of the file server side.
+    url: str
+    #: Size of the file in MB.
+    size: float
+    #: Last modified date.
+    last_modified: str
```
uriii3, line 29: Although it is a string, wouldn't it be better to maintain the DateTime till the end? Maybe also a little bit less confusing for the users, as the date will always be returned in the same format, no? so to not have a "date" with string format (could be anything?), but rather having a closed format (although still a string).

copernicusmarine/core_functions/models.py
```diff
@@ -15,3 +18,73 @@
...
+
+    minimum: Optional[float]
+    maximum: Optional[float]
+
+
+class TimeExtent(BaseModel):
+    """Interval for time coordinate."""
+
+    minimum: Optional[str]
+    maximum: Optional[str]
```
uriii3, line 52: I guess same here

uriii3 (approved):
Looks amazing to me, well done and nice work!

copernicusmarine/catalogue_parser/models.py
```diff
@@ -1,4 +1,6 @@
 import re
+
+# TODO: change to pydantic
```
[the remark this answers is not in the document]
renaudjester: Yes :D It's a task in itself but if one day we want we can

PR #131, opened 2024-09-12
Describe contains coordinate not on the variable

renaudjester:
Fix: [CMT-103](https://cms-change.atlassian.net/browse/CMT-103)

Uses the fact that if the coordinate is not in the chunkLen dict then it means that the variable doesn't have the coordinate

I think the real information is in cube:variables: but I don't want to move around a property

See https://stac.marine.copernicus.eu/metadata/GLOBAL_MULTIYEAR_PHY_ENS_001_031/cmems_mod_glo_phy-all_my_0.25deg_P1D-m_202311/dataset.stac.json for reference


copernicusmarine/catalogue_parser/models.py
```diff
@@ -236,6 +236,7 @@ def from_metadata_item(
                     metadata_item.properties.get("admp_valid_start_date"),
                 )
                 for dimension, dimension_metadata in dimensions.items()
+                if variable_id in dimension_metadata.get("chunkLen", {})
```
uriii3, line 239: but the `cube:variables` is being called right up from this line, no? in line 220. Should the information be there? If not, I'm also fine with using the `chunkLen` option although seems less intuitive when rereading the code.

tests/__snapshots__/test_command_line_interface.ambr
renaudjester, line 1: Maybe it's a big file but I feel like it's nice to have a snapshot of some dataset so we could see some changes

uriii3: mm okay, looks good then!

uriii3 (approved):

PR #132, opened 2024-09-13
Adapt type of valid minmax

uriii3:
In [Pull 127](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/127) we discovered that the values of `valid_min` and `valid_max` didn't have the same format as the variable, which isn't consistent with the CF convention.

I've created this changes to make sure that the types of the values are correct.

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -449,7 +449,7 @@ def _update_variables_attributes(
         dataset[variable].attrs = _filter_attributes(
             dataset[variable].attrs, NETCDF_CONVENTION_VARIABLE_ATTRIBUTES
         )
-        if "valid_min" in dataset[variable].attrs:
+        if "valid_min" and "valid_max" in dataset[variable].attrs:
             _adequate_dtypes_of_valid_minmax(dataset, variable)
     return dataset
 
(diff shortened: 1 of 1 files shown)


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -430,13 +430,27 @@ def _get_variable_name_from_standard_name(
...
+
+
 def _update_variables_attributes(
     dataset: xarray.Dataset, variables: List[str]
 ) -> xarray.Dataset:
     for variable in variables:
         dataset[variable].attrs = _filter_attributes(
             dataset[variable].attrs, NETCDF_CONVENTION_VARIABLE_ATTRIBUTES
         )
+        if "valid_min" and "valid_max" in dataset[variable].attrs:
```
renaudjester, line 452: This is True even is valid_min is not present:

```
In [9]: l = ["valid_max"]

In [10]: "valid_min" and "valid_max" in l
Out[10]: True
```

renaudjester (approved):

copernicusmarine/download_functions/subset_xarray.py
uriii3: Mmh uops :grimacing:
I'll correct it now

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -430,13 +430,30 @@ def _get_variable_name_from_standard_name(
...
     dataset: xarray.Dataset, variables: List[str]
 ) -> xarray.Dataset:
     for variable in variables:
         dataset[variable].attrs = _filter_attributes(
             dataset[variable].attrs, NETCDF_CONVENTION_VARIABLE_ATTRIBUTES
         )
+        if (
+            "valid_min" in dataset[variable].attrs
+            and "valid_max" in dataset[variable].attrs
+        ):
```
uriii3, line 455: better, no? @renaudjester

PR #133, opened 2024-09-13
refactor: use threading for concurrent requests

renaudjester:
Gets rid of aiohttp and nest-async

fix: [CMT-88](https://cms-change.atlassian.net/browse/CMT-88), [CMT-73](https://cms-change.atlassian.net/browse/CMT-73), [CMT-94](https://cms-change.atlassian.net/browse/CMT-94)


copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -49,87 +41,63 @@
...
-        if dataset_id in dataset_metadata_link.href
-    ]
-    if not datasets_metadata_links:
-        return None
-    # TODO: check if we gain a lot of time by doing a gather here
-    # if not don't forget to add a retry policy
-
-    dataset_jsons: list[dict] = await asyncio.gather(
-        *[
+    with CatalogParserConnection() as connection:
```
uriii3, line 68: this also helps to solve some of the warnings in [here](https://github.com/Deltares/dfm_tools/issues/803), no?

renaudjester: Maybe, let's check the warnings when we merged everything

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -285,11 +257,11 @@ def _download_header(
     only_list_root_path: bool = False,
     overwrite: bool = False,
 ) -> Optional[
-    Tuple[
-        Tuple[str, str],
+    tuple[
+        tuple[str, str],
```
uriii3, line 261: so we are now using the default version of tuple from python?

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -539,61 +511,43 @@ def _get_file_size_and_last_modified(
...
 
 
-def _download_files(
-    tuple_original_files_filename: Tuple[
-        str, str, str, list[str], list[pathlib.Path]
-    ],
-) -> list[pathlib.Path]:
-    (
-        username,
+def _original_files_file_download(
```
uriii3, line 514: weird name for the function, no? or do you refer to all the original files from the file requested?

pyproject.toml
```diff
@@ -20,7 +19,6 @@ dask = ">=2022"
 netCDF4 = ">=1.5.4"
 boto3 = ">=1.25"
 semver = ">=3.0.2"
-nest-asyncio = ">=1.5.8"
```
uriii3, line 23: hehe <3

uriii3:
have you noticed or check the behaviour time wise? does it grow the time used or is it more or less the similar? maybe even less? if it increases, is it worth to be able to get rid of nest-asyncio?

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -49,87 +41,63 @@
...
         ) as response:
-            return await response.json()
+            return response.json()
 
-    async def close(self) -> None:
-        await self.session.close()
+    def __enter__(self):
+        return self
 
-    def _retry_policy(self, info: RetryInfo) -> RetryPolicyStrategy:
```
uriii3, line 76: are we getting rid of the retries or is it handled internally somewhere else? the errors seem to be handle another place, but does this mean that we have less control on the behaviour of them?

renaudjester: We are using the `requests` error retry directly 

Yeah I guess a bit less control 🤔

renaudjester:
> have you noticed or check the behaviour time wise? does it grow the time used or is it more or less the similar? maybe even less? if it increases, is it worth to be able to get rid of nest-asyncio?

Yeah I have run some tests and it was very similar times :D

copernicusmarine/download_functions/download_original_files.py
renaudjester: Yeah I like it better I think it's easier than importing the typing hint

copernicusmarine/catalogue_parser/catalogue_parser.py
uriii3: Well, not that I like it ee, I feel like it's easier then, so good!

PR #137, opened 2024-09-17
chore: coordinate selection method

uriii3:
In this PR I will implement the later changes proposed in [CMT-28](https://cms-change.atlassian.net/browse/CMT-28):

We’d like to discuss the following option:
--coordinates-selection-method [inside | strict-inside | nearest | outside]
**Inside Behaviour**: This is the default mode. The requested dimension values must be within the dataset's extent and the selection will be made with an "exact match," as described by the slice of a .sel in xarray.
**Strict-Inside Behaviour**: This is the "error-free" mode. The requested dimension values must be strictly within the dataset's extent (otherwise, an error is displayed with the actual dimension extents), and the selection will be made with an "exact match," as described by the slice of a .sel in xarray.
**Nearest Behaviour**: The requested dimension values can be outside the dataset's extent (in this case, a warning is displayed with the actual dimension extents), but the returned values will be the closest to what is actually contained in the dataset.
**Outside Behaviour**: The requested dimension values can be outside the dataset's extent (in this case, a warning is displayed), and the returned values may include additional points if they exist, based on the dataset's grid.

Let's go step by step:
- [x] Delete bounding-box-method and subset-method from the code
- [ ] Write the ReadMe so that is is understandable -> to do afterwards
- [x] Refactor the tests so that they adapt to the new functionality
- [x] Make all the corrections into the code so that it works
- [ ] Renaud reviews it


uriii3:
should I update the description of the branch?

copernicusmarine/command_line_interface/group_subset.py
```diff
@@ -439,22 +430,21 @@ def subset(
...
+        output_directory=output_directory,
+        credentials_file=credentials_file,
+        motu_api_request=motu_api_request,
+        force_download=force_download,
+        overwrite_output_data=overwrite_output_data,
+        dry_run=dry_run,
+        disable_progress_bar=disable_progress_bar,
+        staging=staging,
+        netcdf_compression_enabled=netcdf_compression_enabled,
+        netcdf_compression_level=netcdf_compression_level,
```
renaudjester, line 447: if you wanna put this in `kwargs` put also the other ones from above

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -146,30 +143,38 @@ def _dataset_custom_sel(
     dataset: xarray.Dataset,
     coord_type: Literal["latitude", "longitude", "depth", "time"],
     coord_selection: Union[float, slice, DateTime, None],
-    bounding_box_method: Optional[Union[BoundingBoxMethod, None]],
-    method: Union[str, None] = None,
+    coordinates_selection_method: Optional[
+        Union[CoordinatesSelectionMethod, None]
+    ],
```
renaudjester, line 148: You don't need the `Union[..,None]` since you have the `Optional` I think

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -146,30 +143,38 @@ def _dataset_custom_sel(
...
                     and coord_selection.stop is not None
                 ):
                     coord_selection = _nearest_selection(
                         dataset, coord_label, coord_selection
                     )
-            tmp_dataset = dataset.sel(
-                {coord_label: coord_selection}, method=method
-            )
+            # això m'ho hauré de mirar bé, cadascun tenia la seva forma
+            # de cridar-lo i no sé si ho podré fer així
```
renaudjester, line 169: Comment: have you checked this in the end?

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -345,21 +341,20 @@ def _temporal_subset(
             if start_datetime == end_datetime
             else slice(start_datetime, end_datetime)
         )
-        temporal_method = "nearest" if start_datetime == end_datetime else None
+        # temporal_method = "nearest" if start_datetime == end_datetime else None
```
renaudjester, line 344: comment

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -408,13 +403,12 @@ def update_elevation_attributes(dataset: xarray.Dataset):
             if minimum_depth == maximum_depth
             else slice(minimum_depth, maximum_depth)
         )
-        depth_method = "nearest" if minimum_depth == maximum_depth else None
+        # depth_method = "nearest" if minimum_depth == maximum_depth else None
```
renaudjester, line 406: comment

med-hcmr-wav-rean-h_VHM0_18.12W-17.00W_38.02N_1993-01-01-1993-01-01.nc
renaudjester, line 1: test file

copernicusmarine/command_line_interface/group_subset.py
uriii3: Seemed more consistent to me!

copernicusmarine/download_functions/subset_xarray.py
uriii3: True!

copernicusmarine/download_functions/subset_xarray.py
uriii3: Yes, the way it worked before was separating the single point selection from the slice selection, setting the 'None' option if it was only one point... it should not happen no more, but I am a little scared (as I was when I actually changed also this part before) that I might not cover all of the cases. Haven't thought of any of it that might not be working, so I am confident with it.

copernicusmarine/command_line_interface/group_subset.py
```diff
@@ -424,37 +415,38 @@ def subset(
...
-        force_download,
-        overwrite_output_data,
-        dry_run,
-        disable_progress_bar,
-        staging,
-        netcdf_compression_enabled,
-        netcdf_compression_level,
+        dataset_id=dataset_id,
+        force_dataset_version=dataset_version,
+        force_dataset_part=dataset_part,
```
uriii3, line 420: what do you think of this? in the `core_functions/subset.py` the part and version are defined like this, maybe if we deprecate the options it makes sense to remove this here, no?

renaudjester: So we deprecated `force_dataset_version` for the user
But for our code internally we want to keep `force_dataset_version` because I think its clearer

renaudjester (approved):

PR #138, opened 2024-09-18
fix: handle admp valid start index for retention period

renaudjester:
We would set the minimun to the `admp_valid_start_date` whereas some datasets have values and hence we should return only the valid values for the time values. See an example: 

https://stac.marine.copernicus.eu/metadata/OCEANCOLOUR_ATL_BGC_L4_NRT_009_116/cmems_obs-oc_atl_bgc-pp_nrt_l4-multi-1km_P1M_202311/dataset.stac.json

### DIFF
--- a/copernicusmarine/catalogue_parser/models.py
+++ b/copernicusmarine/catalogue_parser/models.py
@@ -133,20 +133,30 @@ class CopernicusMarineCoordinate:
         dimension: str,
         dimension_metadata: dict,
         arco_data_metadata_producer_valid_start_date: Optional[str],
+        arco_data_metadata_producer_valid_start_index: Optional[int],
     ) -> Coordinate:
         coordinates_info = dimension_metadata.get("coords", {})
-        if (
-            arco_data_metadata_producer_valid_start_date
-            and dimension == "time"
-        ):
-            minimum_value = (
-                CopernicusMarineCoordinate._format_admp_valid_start_date(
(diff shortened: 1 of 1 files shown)


copernicusmarine/catalogue_parser/models.py
```diff
@@ -133,31 +133,41 @@ def from_metadata_item(
...
+                ]
         chunking_length = dimension_metadata.get("chunkLen")
         if isinstance(chunking_length, dict):
             chunking_length = chunking_length.get(variable_id)
 
         coordinate = cls(
             coordinate_id=dimension,
             units=dimension_metadata.get("units") or "",
-            minimum_value=minimum_value,  # type: ignore
+            minimum_value=minimum_value or coordinates_info.get("min"),  # type: ignore
```
uriii3, line 167: I don't really understand this: if there is a `minimum_value` assigned (not None) it takes it, else it takes `coordinates_info.get("min")`? What happens if there is no `coordinates_info.get("min")`? It throws an error, no?

uriii3: It wouldn't enter the if and then it would have, at max, a `None` as minimum value.

renaudjester: It wouldn't throw an error only `coordinates_info.get("min")` whatever the value is

uriii3 (approved):

Issue #139, opened 2024-09-18
Add logging about where to register for credentials


veenstrajelmer:
When using copernicusmarine in Python, it is not clear which user/pass to provide. This is especially an issue if copernicusmarine is embedded in a larger workflow. If as part of that workflow the `copernicusmarine.login()` function is called and the user is not yet logged in, they will get:
```
username: 
```

It would be great if this is preceded by a line like `"Downloading CMEMS data requires a Copernicus Marine username and password, sign up for free at: https://data.marine.copernicus.eu/register."`

uriii3:
Mmh, yes, it would actually make sense! Thanks for the recommendation!

uriii3:
Thanks for all the feedback, closing this one!

veenstrajelmer:
Thanks a lot for following up on this issue so quick, amazing! I do notice it has been merged into the v2 branch and not in main. What is the timeline for v2 to be released? Just for some expectation management.

uriii3:
Yeees! We are planning to do a pre-release next week. The main release for v2 will be October-November but time is always a bit relative.
Hope this helps!

PR #140, opened 2024-09-18
chore: proposal to update the terminal output for the login

uriii3:
Trying to followup on [Issue-139](https://github.com/mercator-ocean/copernicus-marine-toolbox/issues/139). Seems like a good idea to point out where the registration is from within the toolbox. 
Not sure if the option of having it in the ReadMe is enough, to be discussed.


copernicusmarine/core_functions/credentials_utils.py
renaudjester, line 1: Make a little test

copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -91,6 +91,12 @@ def _load_credential_from_motu_configuration_file(
 def _retrieve_credential_from_prompt(
     credential_type: Literal["username", "password"], hide_input: bool
 ) -> str:
+    if credential_type == "username":
+        logger.info(
+            "Downloading CMEMS data requires a Copernicus Marine username "
+            "and password, sign up for free at:"
+            " https://data.marine.copernicus.eu/register."
```
renaudjester, line 98: without the "."

tests/test_command_line_interface.py
```diff
@@ -831,6 +831,11 @@ def test_login_is_prompt_when_configuration_file_doest_not_exist(
             command, env=environment_without_crendentials
         )
         assert self.output.returncode == 1
+        assert (
+            b"Downloading CMEMS data requires a Copernicus Marine username "
+            b"and password, sign up for free at:"
+            b" https://data.marine.copernicus.eu/register"
+        ) in self.output.stderr
         assert b"username:" in self.output.stdout
```
renaudjester, line 839: check that "copernicusmarine username:" is in the stdout
and maybe the password too 🤔

uriii3: I'm not sure we are entering anything, so not that sure that "coperniscusmarine password" appears there.... but I'll try. Do you prefer that I enter something to check if not?

renaudjester: Ah yeah true :D You can try by putting an input in the `execute_command`

uriii3: check the tests and where they failed, it is this! We don't check that the password exists.

renaudjester (approved):

veenstrajelmer:
Fixes #139 (added this to make sure the PR is linked to the issue)

PR #141, opened 2024-09-18
new-feature: create binaries with GitHub Actions

uriii3:
Add GitHub Actions so that the binaries are created at every release or pre-release.

- [ ] MacOS binary 
- [ ] Linux binary
- [ ] Windows binary
- [ ] Make sure the imports are correct
- [ ] Make sure the `--add-data` option works
- [ ] See where it saves the binaries
- [ ] Maybe also add tests to make sure the binaries work properly


uriii3:
It is an old branch from main as it seems :(

PR #142, opened 2024-09-18
new-feature: create binaries with GitHub Actions 

uriii3:
Add GitHub Actions so that the binaries are created at every release or pre-release.

- [x] MacOS binary 
- [x] Linux binary
- [x] Windows binary
- [x] Make sure the imports are correct
- [x] Make sure the `--add-data` option works
- [x] See where it saves the binaries
- [ ] Maybe also add tests to make sure the binaries work properly


.github/workflows/binaries.yml
```diff
@@ -0,0 +1,57 @@
...
+    branches:
+       - main
+       - 'pre-releases/**'
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, macos-latest, ubuntu-latest]
```
renaudjester, line 14: I thought you wanted 4 binaries (two for mac)?

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: |
+        make run-using-pyinstaller-${{ matrix.os }}
+
+    - name: Upload binaries to windows
+      env:
+        GH_TOKEN: ${{ github.token }}
+      if: matrix.os == 'windows-latest'
+      run: |
+          $MY_VARIABLE = poetry version --short
+          gh release upload v$MY_VARIABLE dist/cmt.exe#${{matrix.os}}-copernicus.exe
```
renaudjester, line 45: `${{matrix.os}}-copernicus.exe` -> `copernicusmarine-${{matrix.os}}.exe`

I think better to have it like this

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,57 @@
...
+        GH_TOKEN: ${{ github.token }}
+      if: matrix.os == 'macos-latest'
+      run: |
+          gh release upload v${{steps.set-version.outputs.VERSION}} dist/cmt.cli#${{matrix.os}}-copernicus
+    - name: Upload binaries to ubuntu
+      env:
+        GH_TOKEN: ${{ github.token }}
+      if: matrix.os== 'ubuntu-latest'
+      run: |
+          gh release upload v${{steps.set-version.outputs.VERSION}} dist/cmt-ubuntu.cli#${{matrix.os}}-copernicus
```
renaudjester, line 57: They are the same no? Maybe you can put then together

.github/workflows/binaries.yml
renaudjester: Same for the other ones

Makefile
```diff
@@ -91,6 +104,25 @@ build-and-publish-dockerhub-image:
 	docker push copernicusmarine/copernicusmarine:$${VERSION}
 	docker push copernicusmarine/copernicusmarine:latest
 
+build-and-prepare-for-binary-macos-latest:
+	python -m pip install --upgrade pip
+	pip install pyinstaller
+	pip install -e .
+	pip install poetry
```
renaudjester, line 111: you can take this part and put it into a make command

renaudjester: I still think that it would be better to do all of this into a mamba env

renaudjester: Also with the mamba env it seems like you have access to a power-shell for windows

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Build with PyInstaller
+      run: |
+        make run-using-pyinstaller-${{ matrix.os }}
+
+    - name: Upload binaries to windows
+      env:
+        GH_TOKEN: ${{ github.token }}
+      if: matrix.os == 'windows-latest'
+      run: |
+          $MY_VARIABLE = poetry version --short
```
renaudjester, line 44: Maybe rename the variable to `VERSION` 

Also this doesn't work for Ubuntu and Mac? If yes maybe you can just use it for all

.github/workflows/binaries.yml
uriii3: oh, yes!

Makefile
uriii3: It is already into a makefile, no? I don't understand what you mean.

Yes, but I was a bit scared because I don't fully understand the mamba things so I'm not so sure that it would work better... seems to work fine like this!

.github/workflows/binaries.yml
uriii3: Mmh haven't check if it works or not (it is in PowerShell), but the main difference with Mac and ubuntu is that the executable in windows needs to have the `.exe` into the name, so then the following part of the command would need to be different anyway!

Makefile
renaudjester: Yeah I know the mamba thing can be some work but it would be nice because it keeps environment installation the same across the repo
In the repo, for github action we never use pure pip commands

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,58 @@
...
+        make run-using-pyinstaller-${{ matrix.os }}
+
+    - name: Upload binaries to windows
+      env:
+        GH_TOKEN: ${{ github.token }}
+      if: matrix.os == 'windows-latest'
+      run: |
+          $WIN_VERSION = poetry version --short
+          gh release upload v$WIN_VERSION dist/cmt.exe#copernicusmarine-windows.exe
+    - name: Upload binaries to macos
```
renaudjester, line 46: Can be renamed to `Upload binaries to macos or ubuntu`

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+(describe or any other command)
+
+You might have to update the permissions of the binary to be able to execute it with linux:
+
+.. code-block:: bash
+
+    chmod +rwx cmt_ubuntu-latest.cli
+
+And from a Windows os:
```
renaudjester, line 63: Add where you open the shell: powershell or else?

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+Use the CopernicusMarine Toolbox binaries
+***********************************************
+
+In the release page (https://github.com/mercator-ocean/copernicus-marine-toolbox/releases) you can access the binaries of the latest releases. Once downloaded for the specific platform, you can use the toolbox by running the binary as follows:
+
+In mac-os or linux:
+
+.. code-block:: bash
+
+    ./cmt_macos-latest.cli describe
```
renaudjester, line 53: The names are not correct right?

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. code-block:: bash
+
+    chmod +rwx cmt_ubuntu-latest.cli
+
+And from a Windows os:
+
+.. code-block:: bash
+
+    cmt_windows-latest.exe describe
```
renaudjester, line 67: Same here since you rename it

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
+===================================================
+Installing Copernicus Marine Toolbox
+===================================================
+
+There are several ways to install or use the Copernicus Marine Toolbox:
+
+* via pip `pypi`_
+* via conda `conda-forge`_
+* via docker `dockerhub`_
+
```
renaudjester, line 10: Add also maybe a line: 
`Alternatively you can use a binary.`

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+    chmod +rwx cmt_ubuntu-latest.cli
+
+And from a Windows os:
+
+.. code-block:: bash
+
+    cmt_windows-latest.exe describe
+
+(describe or any other command)
```
renaudjester, line 69:
```suggestion
(``describe`` or any other command)
```

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+In the release page (https://github.com/mercator-ocean/copernicus-marine-toolbox/releases) you can access the binaries of the latest releases. Once downloaded for the specific platform, you can use the toolbox by running the binary as follows:
+
+In mac-os or linux:
+
+.. code-block:: bash
+
+    ./cmt_macos-latest.cli describe
+
+(describe or any other command)
```
renaudjester, line 55:
```suggestion
(``describe`` or any other command)
```

doc/usage/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. _pypi: https://pypi.org/project/copernicusmarine/
+.. _conda-forge: https://anaconda.org/conda-forge/copernicusmarine
+.. _dockerhub: https://hub.docker.com/r/copernicusmarine/copernicusmarine
+.. _copernicusmarine-register: https://data.marine.copernicus.eu/register
+
+Use the CopernicusMarine Toolbox binaries
+***********************************************
+
+In the release page (https://github.com/mercator-ocean/copernicus-marine-toolbox/releases) you can access the binaries of the latest releases. Once downloaded for the specific platform, you can use the toolbox by running the binary as follows:
```
renaudjester, line 47:
```suggestion
In the `release page <https://github.com/mercator-ocean/copernicus-marine-toolbox/releases>`_ you can access the binaries of the latest releases. Once downloaded for the specific platform, you can use the toolbox by running the binary as follows:
```

Makefile
```diff
@@ -54,6 +54,21 @@ run-tests-dependencie-versions:
 	${ACTIVATE_ENVIRONMENT}
 	tox run
 
+run-using-pyinstaller-windows-latest:
+	pip install distributed
```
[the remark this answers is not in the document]
uriii3: We discussed it with Gery the other day, it is not that clear (to me) that the different steps are reproduced in the same environment, which is quite weird to me but looked that way. So I think it is safer to keep it here.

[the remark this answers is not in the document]
renaudjester: I don't get what you mean?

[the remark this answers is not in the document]
renaudjester: from the moment you have the matrix all the steps will be on the different environment of the matrix value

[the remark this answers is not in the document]
uriii3: Mmh maybe you are right, let me try it! I think we saw something about different environments in deferent 'steps' but on the same machine, which seemed quite weird but that was our logic. I don't really remember now.

doc/usage/installation.rst
uriii3: Uuh okey, is a cmd. Would it be better to also wrote into PowerShell?

doc/usage/installation.rst
uriii3: actually, they are :( I upload them in a different name (the one that appears in the release section) but when you download, all of them recover their original name (this ones to be exactly). Should we change them or make them a little bit more explicit?

doc/usage/installation.rst
renaudjester: No but just say it in the documentation: `Opening a cmd command prompt on windows`

uriii3: Mmh okay, done then!

doc/usage/installation.rst
renaudjester: Ah ok :D well nice! then you can name the assets with more human readable names line: `copernicusmarine binary for windows v2.0.0` 
and then change the name of the file to be a bit prettier

doc/usage/installation.rst
```diff
@@ -0,0 +1,70 @@
...
+Use the CopernicusMarine Toolbox binaries
+***********************************************
+
+In the `release page <https://github.com/mercator-ocean/copernicus-marine-toolbox/releases>`_ you can access the binaries of the latest releases. Once downloaded for the specific platform, you can use the toolbox by running the binary as follows:
+
+In mac-os or linux:
+
+.. code-block:: bash
+
+    ./cmt_macos-latest.cli describe
```
renaudjester, line 54: change the names

doc/usage/installation.rst
```diff
@@ -0,0 +1,70 @@
...
+
+.. code-block:: bash
+
+    chmod +rwx cmt_ubuntu-latest.cli
+
+And from a Windows os (cmd):
+
+.. code-block:: bash
+
+    cmt_windows-latest.exe describe
```
renaudjester, line 68: Now they are outdated right?

doc/usage/installation.rst
```diff
@@ -0,0 +1,70 @@
+===================================================
+Installing Copernicus Marine Toolbox
+===================================================
+
+There are several ways to install or use the Copernicus Marine Toolbox:
+
+* via pip `pypi`_
+* via conda `conda-forge`_
+* via docker `dockerhub`_
+* alternatively, you can use a binary
```
renaudjester, line 10: Maybe don't put it in the list

Makefile
```diff
@@ -54,6 +54,19 @@ run-tests-dependencie-versions:
...
 	tox run
 
+run-using-pyinstaller-windows-latest:
+	python -m PyInstaller --copy-metadata xarray --name copernicusmarine-toolbox.exe --add-data "c:\hostedtoolcache\windows\python\3.12.6\x64\lib\site-packages\distributed\distributed.yaml;.\distributed" copernicusmarine/command_line_interface/copernicus_marine.py --onefile
+
+run-using-pyinstaller-macos-latest:
+	python -m PyInstaller --name copernicusmarine-toolbox_macos-latest.cli copernicusmarine/command_line_interface/copernicus_marine.py --onefile --target-architecture=arm64
+
+run-using-pyinstaller-macos-13:
+	python -m PyInstaller --name copernicusmarine-toolbox_macos-13.cli copernicusmarine/command_line_interface/copernicus_marine.py --onefile --target-architecture=x86_64
```
renaudjester, line 64: don't name them based on the name of the github machines. Maybe just name them all the same: `copernicusmarine.cli` (not sure about the toolbox, we can ask later for this details)

uriii3: completely true!

uriii3: Nope, I remember why I did this. If they all have the same name, you can not upload them all in the release as different types of assets, as GitHub reads them all as the same (they have the same name).

renaudjester: Ok then change the name to be where they are suppose to work

renaudjester: Like copernicusmarine-toolbox_linux
copernicusmarine-toolbox_macos-arm64 
etc

renaudjester (approved):

Makefile
uriii3: meh but then it fucks up my automation system: the name `copernicusmarine_ubuntu-latest.cli` is good because it makes all the calls within one line: `ist/copernicusmarine_${{matrix.os}}.cli`. If it doesn't, it is a bit more difficult to automate (maybe I can get rid of the 'latest' as if it was a vector from python...)
I can adapt it if you want to, but might not be trivial.

uriii3: Or at least it is not straight forward

PR #143, opened 2024-09-18
fix: subset should subset when retention period

renaudjester:
The retention period needs to be applied when doing a subset, hence we need to handle the case where we only have values for the time coordinate. 

Was missing to this [PR](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/138).


copernicusmarine/core_functions/services_utils.py
```diff
@@ -439,10 +439,16 @@ def _get_dataset_start_date_from_service(
...
         for coordinate in variable.coordinates:
-            if coordinate.coordinate_id == "time" and coordinate.minimum_value:
-                if isinstance(coordinate.minimum_value, str):
-                    return coordinate.minimum_value.replace("Z", "")
-                return int(coordinate.minimum_value)
+            if coordinate.coordinate_id == "time":
+                if coordinate.minimum_value:
+                    if isinstance(coordinate.minimum_value, str):
+                        return coordinate.minimum_value
+                    return int(coordinate.minimum_value)
```
uriii3, line 446: sometimes I get lost here, because we are considering that if it is a string, we return a string, if it is not a string (shouldn't it be a int, then?) we return an integer? so we keep the type of the original format, no?
Why then the `if` statement?

uriii3 (approved):
so basically what we are assuming is that dates can also be floats? isn't it a little weird?

renaudjester:
> so basically what we are assuming is that dates can also be floats? isn't it a little weird?

No because a timestamp for time can be a float too

PR #146, opened 2024-09-23
feat: add timeout to requests and can choose the retry number

### LINKED ISSUE #109: Missing timeout
## Motivation

In our application https://opendrift.github.io/ copernicusmarine is used as one of several sources. This bug causes the application to stall instead of using other sources.

## Expected behavior

A default timeout should be set, and should be overridable.

## Actual behavior

Copernicusmarine doesn't set timeout for network requests. When the object store is down, this causes the module to wait forever. 

## Steps to reproduce

- Open netcat to listen on a port, but not respond: sudo nc -v -l 443
- Point all object store addresses to this machines IP, using the hosts file.

> $ tail /etc/hosts
> 157.249.79.36 s3.waw3-1.cloudferro.com
> 157.249.79.36 cloudferro.com
> 157.249.79.36 *.cloudferro.com

Run `copernicusmarine describe` and notice that it will wait indefinitely.

## Environment

Operating System: Ubuntu 22.04 64-bit
Python version: Python 3.10.12
copernicusmarine version: 1.2.4
Source execution: terminal
RAM: 32 GB
Bandwidth: Gigabit network
Use of proxy: no
Use of VPN: no
Date: There was a cloud outage 06.08.2024 which surfaced these problems.
Network Permission: No limitations on network.

renaudjester:
Fix [CMT-101](https://cms-change.atlassian.net/browse/CMT-101)
Closes #109 
and closes #110


doc/usage/network-configuration.rst
```diff
@@ -58,3 +58,16 @@ The default value is ``15`` and minimum value is ``1``.
...
+Connection Timeout and Retries
+-------------------------------
+
+The toolbox uses the ``requests`` library to make HTTP calls. By default, the connection timeout is set to 60 seconds.
+It's the timeout argument that is being set. Please refer to `requests documentation on timeouts <https://requests.readthedocs.io/en/stable/user/advanced/#timeouts>`_
+for more information.
+
+You can set the ``COPERNICUSMARINE_HTTPS_TIMEOUT`` environment variable to a custom value in seconds.
+The default value is ``60``. Accepted values are positive floats.
+
```
uriii3, line 71: maybe it would be interesting to show a way or 2 to set this environment variable (it is not straightforward for the users maybe...)

copernicusmarine/core_functions/sessions.py
```diff
@@ -69,29 +79,33 @@ def get_configured_boto3_session(
...
+# https://medium.com/@mpuig/testing-robust-requests-with-python-a06537d97771
+class ConfiguredRequestsSession(requests.Session):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.trust_env = TRUST_ENV
+        self.verify = (
+            COPERNICUSMARINE_SET_SSL_CERTIFICATE_PATH or certifi.where()
+        )
+        self.proxies = PROXIES
+        if HTTPS_RETRIES:
```
uriii3, line 93: what if they put retries to 0? so they don't want to have any retries?

uriii3 (approved):
I've checked the requests.Session page and it looks okay from what I understood! with the way you create the session I think it will also be better, so hope it works better now!

doc/usage/network-configuration.rst
renaudjester: There are some examples in the documentation I think :D

copernicusmarine/core_functions/sessions.py
renaudjester: Yep

PR #147, opened 2024-09-23
chore: clean repo and delete deprecated

renaudjester:
For the v2 we can get rid for the deprecated things

- [x]  Add warning that version >2 of numpy can cause problems with older versions of xarray


copernicusmarine/core_functions/versions_verifier.py
```diff
@@ -61,6 +61,7 @@ def _check_version(function_name: str, staging: bool):
                     f"Using a pre-release or a non-official version "
                     f"of the client. Client version: {client_version}"
                 )
+                return
```
uriii3, line 64: what is this about?

renaudjester: Just that before it would print twice because we are in a loop so I am just returning to print it once

PR #149, opened 2024-09-25
fix: correcting binaries docs

uriii3:
Updating the explanation of the binaries where they belong, so that the docs are properly seen.


doc/usage/installation.rst
renaudjester, line 1: I think you can delete 🤔

renaudjester (approved):

PR #150, opened 2024-09-26
upgrade the binaries GitHub Action

uriii3:
Upgrade the binaries GitHub Action so that it uses the micro mamba shell, for consistency and for better coding.


uriii3:
- [ ] I've separated windows for other os (`micromamba-binary-path` is differently set in both cases and so it was complicating all the workflow too much).

- [ ] Some of the pyinstaller now need additional data packages, but that seems fine for me, just to make sure they work.

- [ ] I've checked the binaries for linux and windows in the VM's to make sure they work, it is a pass in them! 

- [ ] I also checked with Mac, it seems like it is a bit reluctant to open the executable if it comes from an 'unknown' developer... still, might be something to look at? Let's see what the feedback is.

- [ ] I also created the logos for the executables, although it only seems to work in windows for now. I am a bit unsure on where to store the image, if you have any recommendation it will be welcomed.

renaudjester (approved):

PR #153, opened 2024-10-01
fix: update create template

uriii3:
We discovered that with this [PR](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/72), where we added a warning also in the depth dimension, now the option `--create-template` for the subset wasn't working properly.

To solve it, I changed some of the values provided in the create-template option.

I also added some tests, although as with the `--create-template` option there is no other valid inputs, I don't know if the template will be saved some place I can recover easily from the tests. We will see now.


copernicusmarine/core_functions/subset.py
```diff
@@ -218,8 +218,8 @@ def create_subset_template() -> None:
                 "maximum_longitude": -10,
                 "minimum_latitude": 35,
                 "maximum_latitude": 43,
-                "minimum_depth": False,
-                "maximum_depth": False,
+                "minimum_depth": 0,
```
renaudjester, line 221: Do you have a warning due to this? If yes, maybe better not to have a warning in the template

uriii3: Yes! The `dataset_id` actually is a bit weird and contains only one point (0.49...), so the only way to get rid of that would be to change the dataset or put the exact point.
I would maybe go for the change in the `dataset_id`, if you prefer I don't move that say so too.

renaudjester (approved):

PR #154, opened 2024-10-03
chore: update the output when listing files

uriii3:
From ticket [CMT-108](https://cms-change.atlassian.net/browse/CMT-108), were a get was being done and the toolbox was taking so long, we thought that it might be interesting to add, as an output, some sort of information about the process that is being done underneath.

This might help users understand that, although it will take a while, it will eventually reach and download the files.

### DIFF
--- a/copernicusmarine/download_functions/download_original_files.py
+++ b/copernicusmarine/download_functions/download_original_files.py
@@ -467,7 +467,7 @@ def _list_files_on_marine_data_lake_s3(
         Prefix=prefix,
         Delimiter="/" if not recursive else "",
     )
-    logger.info("Listing files on the server...")
+    logger.info("Listing files on remote server...")
     s3_objects = chain(
         *map(lambda page: page.get("Contents", []), tqdm(page_iterator))
     )
--- a/copernicusmarine/download_functions/download_original_files.py
+++ b/copernicusmarine/download_functions/download_original_files.py
@@ -457,7 +457,6 @@ def _list_files_on_marine_data_lake_s3(
     prefix: str,
     recursive: bool,
 ) -> list[tuple[str, int, DateTime, str]]:
-
     s3_client, _ = get_configured_boto3_session(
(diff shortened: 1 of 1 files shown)


copernicusmarine/download_functions/download_original_files.py
```diff
@@ -468,11 +467,10 @@ def _list_files_on_marine_data_lake_s3(
         Prefix=prefix,
         Delimiter="/" if not recursive else "",
     )
-
+    logger.info("Listing files on the server...")
```
renaudjester, line 470:
```suggestion
    logger.info("Listing files on remote server...")
```

renaudjester (approved):

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -468,11 +467,10 @@ def _list_files_on_marine_data_lake_s3(
         Prefix=prefix,
         Delimiter="/" if not recursive else "",
     )
-
+    logger.info("Listing files on remote server...")
     s3_objects = chain(
-        *map(lambda page: page.get("Contents", []), page_iterator)
+        *map(lambda page: page.get("Contents", []), tqdm(page_iterator))
```
renaudjester, line 472: need to be disabled with the disable progress bar

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -468,11 +475,13 @@ def _list_files_on_marine_data_lake_s3(
         Prefix=prefix,
         Delimiter="/" if not recursive else "",
     )
-
+    logger.info("Listing files on remote server...")
```
uriii3, line 478: what do you think, should I also delete this print if the flag `disable-progress-bar` is activated? for me it makes more sense maybe, no? to not have anything if they disable the progress bar

uriii3: Mmh okay, alright! I will merge when it passes the tests then!

PR #156, opened 2024-10-07
fix: correct typos and improve doc

renaudjester:
Based on some comments after reading https://toolbox-docs.marine.copernicus.eu/en/v2.0.0a2/


copernicusmarine/command_line_interface/group_subset.py
```diff
@@ -60,23 +60,13 @@ def cli_subset() -> None:
...
-        --variable thetao
-        --start-datetime 2022-01-01T00:00:00
-        --end-datetime 2022-12-31T23:59:59
-        --minimum-longitude -6.17
-        --maximum-longitude -5.08
-        --minimum-latitude 35.75
-        --maximum-latitude 36.30
-        --minimum-depth 0.0
-        --maximum-depth 5.0
+        copernicusmarine subset --dataset-id cmems_mod_ibi_phy_my_0.083deg-3D_P1D-m --variable thetao --variable so --start-datetime 2021-01-01 --end-datetime 2021-01-03 --minimum-longitude 0.0 --maximum-longitude 0.1 --minimum-latitude 28.0 --maximum-latitude 28.1 --minimum-depth 1 --maximum-depth 2
```
uriii3, line 63: weren't they saying also that they didn't like that `..code-block:: bash` was appearing? is this solved on how to call the sphinx util?

renaudjester: You are right I should check this!

renaudjester: Thanks!

doc/installation.rst
```diff
@@ -1,23 +1,27 @@
...
 There are several ways to install or use the Copernicus Marine Toolbox:
 
-* via pip `pypi`_
-* via conda `conda-forge`_
-* via docker `dockerhub`_
+* via pip (see `PyPI repository <https://pypi.org/project/copernicusmarine/>`_)
+* via mamba | conda (see `conda-forge channel <https://anaconda.org/conda-forge/copernicusmarine>`_)
+* via docker (see `dockerhub repository <https://hub.docker.com/r/copernicusmarine/copernicusmarine>`_)
 
 Alternatively, you can use a binary.
```
uriii3, line 11: not for this PR (so, note to myself and you), but should we point somewhere here too? at the end of the section or something like that?

doc/installation.rst
```diff
@@ -1,23 +1,27 @@
...
 
+.. note::
+
+    Requires Python ``>=3.9`` and ``<3.13``.
+
 .. note::
 
     Note that the use of ``xarray<2024.7.0`` with ``numpy>=2.0.0`` leads to inconsistent results. See this issue: `xarray issue <https://github.com/pydata/xarray/issues/9179>`_.
```
uriii3, line 20: how much can we change this? for example: xarray solves it in a new release: we don't release until one month in... will the documentation stay the same until we rerelease? or there is a way to update the documentation without releasing? is it good, then, to have this sort of comments here? (for me it is but maybe not the best way that we don't have a way to update the documentation

doc/usage/describe-usage.rst
```diff
@@ -14,3 +54,72 @@ To save the JSON output to a file, you can use the following command:
 .. code-block:: bash
 
     copernicusmarine describe --include-datasets > all_datasets_copernicusmarine.json
+
```
uriii3, line 57: here there is no response because it will all be sent to a json, but then we are not showing any sort of response... understandable from a logic point of view but maybe from a user side some answer would be more helpful? regarding what they said about obtaining responses always.

doc/command-line-interface.rst
```diff
@@ -1,6 +1,33 @@
-CLI
+.. _command-line-interface:
+
+Command line interface
 ===================================================
 
 .. click:: copernicusmarine.command_line_interface.copernicus_marine:base_command_line_interface
     :prog: copernicusmarine
-    :nested: full
```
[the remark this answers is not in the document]
renaudjester: So by manually adding them instead of letting `sphinx-click` do the nested documentation, it allows to skip some titles that were like `cli-describe` (it was a not from David)

doc/usage/describe-usage.rst
```diff
@@ -14,3 +54,72 @@ To save the JSON output to a file, you can use the following command:
...
+``--contains`` option
+----------------------
+
+You also have the option to filter the output by using the ``--contains`` option. It will perform a search on all the text fields of the output.
+
+**Example:**
+
+If you want the `cmems_obs-ins_glo_phy-temp-sal_my_cora_irr` dataset only, you can use the following command:
```
uriii3, line 65: Maybe: "If you want, **for example,** the `comes...` this way it takes importance of the name of the dataset(?)

doc/installation.rst
renaudjester: But if they solve it in a new release then this statement is still true, isn't it? 
So in which configuration would this become false?

doc/usage/get-usage.rst
```diff
@@ -71,24 +71,25 @@ About sync option
...
 ------------------------
 
 The ``--filter`` option allows specifying a Unix shell-style wildcard pattern to select specific files.
 
-**Example:**
+**Example** To download only files that contains "2000", "2001", or "2002":
 
 .. code-block:: bash
 
     copernicusmarine get --dataset-id cmems_mod_ibi_phy_my_0.083deg-3D_P1Y-m --filter "*01yav_200[0-2]*"
```
uriii3, line 89: the answer of this command too?

doc/usage/get-usage.rst
```diff
@@ -71,24 +71,25 @@ About sync option
...
 
 Option ``--regex`` allows specifying a regular expression for more advanced file selection.
 
-**Example:**
+**Example** To download only files that contains "2000", "2001", or "2002" using a regular expression:
 
 .. code-block:: bash
```
uriii3, line 94: maybe the answer for this one too?

doc/usage/describe-usage.rst
renaudjester: But I already show an example before and after so I thought this was enough :D And about that:

> what they said about obtaining responses always.

It concerns the docstrings, not all the commands of the documentation

doc/usage/describe-usage.rst
renaudjester: it takes importance of the name? what does it mean?

doc/usage/get-usage.rst
renaudjester: I added a little line to describe, I thought this was enough because in the end the response logs are not super relevant

doc/usage/get-usage.rst
renaudjester: cf answer from comment before

doc/usage/login-usage.rst
```diff
@@ -12,11 +12,11 @@ The ``login`` command creates a configuration file called ``.copernicusmarine-cr
     password :
     INFO - Configuration files stored in /Users/foo/.copernicusmarine
 
-If the ``.copernicusmarine-credentials`` file already exists, the system will ask for confirmation before overwriting it (using the ``--overwrite`` or ``--overwrite-configuration-file`` options).
+If the ``.copernicusmarine-credentials`` file already exists, the system will ask for confirmation before overwriting it. You can also use option ``–-overwrite`` or ``--overwrite-configuration-file`` to skip confirmation.
```
uriii3, line 15: this makes my life weirder: we have a flag which is `--skip-if-user-logged-in` because the default behaviour is to precisely overwrite the credentials file... I'm not understanding this option then (or is this a new one which will change in favour of the other? In that case, I would be against this and maybe some check to see if it exists -but not overwrite it!-)

doc/usage/subset-usage.rst
```diff
@@ -71,7 +69,7 @@ About ``--netcdf3-compatible`` option
 """"""""""""""""""""""""""""""""""""""""
 
 The ``--netcdf3-compatible`` option enables compatibility with the netCDF3 format.
-This uses the ``format="NETCDF3_CLASSIC"`` setting in the xarray `to_netcdf` method. `Documentation link <https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_netcdf.html>`_.
+This uses the ``format="NETCDF3_CLASSIC"`` setting in the xarray `to_netcdf` method. (cf. `xarray documentation <https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_netcdf.html>`_.)
 
 About ``--bounding-box-method`` option
 """"""""""""""""""""""""""""""""""""""""
```
uriii3, line 75: I will need to delete this!

doc/installation.rst
uriii3: Mmh if they solve it and then 2024.7.1 works (would be easier for users maybe?) or is it not that big of a priority?

doc/usage/describe-usage.rst
uriii3: that it is more easy to understand that the dataset id is random. It was a suggestion, definitely not super sure about it

doc/usage/login-usage.rst
renaudjester: I still don't understand the `--skip-if-user-logged-in` but concerning the `--overwrite` here is how I understand the process: 

You are a user, you do: `copernicusmarine login`, you log in as "USER1" and a file is created: `~/.../.credentials`. 
Then for some reason (for example you were on a common server) someone comes and wants to log in as "USER2". 
They do `copernicusmarine login` and this will be prompt:
```
copernicusmarine password:
File /../.credentials already exists, overwrite it ? [y/N]:
```
All good :D

Now USER1 and USER2 want to automatise this because they want to launch their scripts but they don't know who is connected. Hence they will use the `--overwrite` argument to avoid needed to prompt anything and let the code run!

doc/usage/shared-options.rst
```diff
@@ -97,15 +97,93 @@ Option ``--credentials-file``
...
+          "maximum": 56.0
+        },
+        "time": {
+          "minimum": "1993-01-01T00:00:00Z",
+          "maximum": "2021-12-28T00:00:00Z"
+        },
+        "depth": {
+          "minimum": 0.5057600140571594,
+          "maximum": 5698.060546875
+        }
```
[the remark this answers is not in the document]
renaudjester: Exactly :D

doc/usage/subset-usage.rst
renaudjester: (it's in the document I shared you I think)

doc/installation.rst
renaudjester: If they solve it then 2024.7.1 works then it's good :D We still have to warn the users that if they use version 2024.7.0 it might bring some bugs with numpy=2.0.0 right?

doc/usage/login-usage.rst
uriii3: Okay, then maybe it is worth to give it two thoughts... 

From what I understand, and following your example: if user1 goes again in the machine, no one has touched it, and wants to login but not do it necessarily if it already has it's credentials, the flag `--skip-if-user-logged-in` precisely would skip the login for user 1.

What seems weird, though, is that this same example would be similar if user2 had logged in in the middle... so user1 would be using the wrong credentials (maybe not a big deal?)

uriii3: Maybe to be thought but outside of the scope of this PR!

uriii3 (approved):

doc/usage/login-usage.rst
renaudjester: Yes we will study this when doing the PR on this subject

PR #157, opened 2024-10-07
fix: no service option for get command

renaudjester:
Get command only have one service available so getting rid of the option to choose it for the user


copernicusmarine/core_functions/get.py
```diff
@@ -154,7 +149,7 @@ def _run_get_request(
         get_request.dataset_id,
         get_request.force_dataset_version,
         get_request.force_dataset_part,
-        get_request.force_service,
+        None,
```
uriii3, line 152: do we actually want to keep this?

renaudjester: I think so because the `get_retrieval_service` is also used by the subset that have the option to choose the service :D

PR #160, opened 2024-10-09
Login prompt

uriii3:
Change and update login prompt as requested


uriii3:
Makes no sense after your [branch](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/161), I close it in favour of yours!

PR #161, opened 2024-10-10
chore: add validity check for logging

renaudjester:
Fix [CMT-110](https://cms-change.atlassian.net/browse/CMT-110)

Allow user to check the validity of their credentials


copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -114,7 +119,7 @@ def _retrieve_credential_from_prompt(
             " https://data.marine.copernicus.eu/register"
         )
     return click.prompt(
-        "copernicusmarine " + credential_type, hide_input=hide_input
+        "Copernicus Marine " + credential_type, hide_input=hide_input
```
uriii3, line 122: I didn't want to commit them, I wanted to make a suggestion, sorry 😨

renaudjester: No problem it was a good idea

copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -206,9 +211,60 @@ def copernicusmarine_configuration_file_exists(
     return configuration_filename.exists()
 
 
+def copernicusmarine_credentials_are_valid(
+    configuration_file_directory: pathlib.Path,
+    username: Optional[str],
+    password: Optional[str],
+):
+    if username and password:
+        if copernicusmarine_username_password_are_valid(username, password):
```
uriii3, line 220: we could jump straight to the function `_check_credentials_with_cas`, no? the username and password will exist for sure here

copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -206,9 +211,60 @@ def copernicusmarine_configuration_file_exists(
...
+        else:
+            logger.info(
+                "Invalid credentials from input username and password."
+            )
+            logger.info(RECOVER_YOUR_CREDENTIALS_MESSAGE)
+            return False
+    elif (
+        COPERNICUSMARINE_SERVICE_USERNAME and COPERNICUSMARINE_SERVICE_PASSWORD
+    ):
+        if copernicusmarine_username_password_are_valid(
```
uriii3, line 232: here too, the options (Environment username and environment password) will exist for sure, no? you are checking just before

copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -206,9 +211,60 @@ def copernicusmarine_configuration_file_exists(
     return configuration_filename.exists()
 
 
+def copernicusmarine_credentials_are_valid(
```
uriii3, line 214: what is the purpose of checking first the username and password? So, if the option exists, is only to check if the user is already logged in, no? So return True if environement variables or credentials are set. Why would there be the option `--check-credentials-valid` if then it inputs username and password?

copernicusmarine/command_line_interface/group_login.py
```diff
@@ -92,18 +98,19 @@ def login(
     password: Optional[str],
     configuration_file_directory: pathlib.Path,
     overwrite_configuration_file: bool,
-    skip_if_user_logged_in: bool,
+    check_credentials_valid: bool,
     log_level: str = "INFO",
 ) -> None:
```
uriii3, line 103: with this option, wouldn't it be normal to expect True or False as output of the login function?

renaudjester: In the CLI returning true or false doesn't change anything 
So I am either returning 0: when well you are connected or 1: there has been an error

copernicusmarine/core_functions/login.py
```diff
@@ -16,19 +16,15 @@ def login_function(
     password: Optional[str],
     configuration_file_directory: pathlib.Path,
     overwrite_configuration_file: bool,
-    skip_if_user_logged_in: bool,
+    check_credentials_valid: bool,
 ) -> bool:
```
uriii3, line 20: okey, this one is actually a bool, makes sense now hehe

copernicusmarine/core_functions/credentials_utils.py
renaudjester: Because that's how the other commands work: if you do a get and you set some username and password, even if you have a correct credential file it will fail. There is priority in the credentials that are taken depending on where they come from

tests/test_command_line_interface.py
```diff
@@ -967,36 +839,8 @@ def test_get_download_no_files(self):
...
-            "login",
-            "--configuration-file-directory",
-            f"{non_existing_directory}",
-            "--skip-if-user-logged-in",
-        ]
-        output_with_skip = execute_in_terminal(command_with_skip)
-        assert output_with_skip.returncode == 0
-
     # TODO: separate tests for each service
-    # SUBSET, GET, LOGIN, DESCRIBE
```
uriii3, line 999: I was being scared like: why did Renaud delete all those tests 😮 then I scrolled down hehe

tests/test_login.py
```diff
@@ -0,0 +1,321 @@
...
+            "--credentials-file",
+            f"{credentials_file}",
+        ]
+
+        self.output = execute_in_terminal(
+            command, env=environment_without_crendentials
+        )
+        assert self.output.returncode == 1
+        assert (
+            b"Downloading CMEMS data requires a Copernicus Marine username "
```
uriii3, line 81: this is a reminder that this output message should be deleted: no CMEMS on the toolbox!

tests/test_login.py
```diff
@@ -0,0 +1,321 @@
...
+            b"and password, sign up for free at:"
+            b" https://data.marine.copernicus.eu/register"
+        ) in self.output.stderr
+        assert b"Copernicus Marine username:" in self.output.stdout
```
uriii3, line 85: Sorry if I broke a test for that

tests/test_login.py
renaudjester: Read the other PR

tests/test_login.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def test_login_command_prompt_with_other_commands(self, tmp_path):
+        self.check_credentials_username_specified_password_prompt(tmp_path)
```
uriii3, line 88: what sense does this make? have a test that invokes another test(?)

copernicusmarine/command_line_interface/group_login.py
```diff
@@ -72,10 +72,16 @@ def cli_login() -> None:
     help="Flag to skip confirmation before overwriting configuration file.",
 )
 @click.option(
-    "--skip-if-user-logged-in",
+    "--check-credentials-valid",
     is_flag=True,
     default=False,
```
uriii3, line 77: wouldn0t it make more sense to make the default as True? So that, by default, we don't overwrite the credentials file if they already exist? -> if they are already connected no need to connect again!

copernicusmarine/core_functions/credentials_utils.py
uriii3: Mmmh but why would you introduce username and password if they already exist? in the login it did pop up always the login when ran, but the get worked autonomously, first checking the environment and credentials file, no? It doesn't ask me my username and password(?)

tests/test_login.py
```diff
@@ -0,0 +1,318 @@
...
+            f"{non_existing_directory}",
+        ]
+
+        self.output = execute_in_terminal(command)
+
+        environment_without_crendentials = (
+            get_environment_without_crendentials()
+        )
+
+        command = [
```
uriii3, line 277: why are we rewriting the definition of the  `command`?

uriii3: and we are using the same 'self' then to assert... makes for a weird test, no?

copernicusmarine/command_line_interface/group_login.py
uriii3: Now that I understand this doesn't make sense!

copernicusmarine/core_functions/credentials_utils.py
uriii3: Now this makes sense too

tests/test_login.py
renaudjester: Because in python you can :D the object is just overwritten and since we don't need both at the same time, why not?

tests/test_login.py
```diff
@@ -0,0 +1,318 @@
...
+
+        assert (folder / ".copernicusmarine-credentials").is_file()
+        assert login(
+            check_credentials_valid=True,
+        )
+
+        assert not login(
+            username="toto",
+            password="tutu",
+        )
```
uriii3, line 318: Wouldn't you add a test specific also for the option that you created? For the python interface a little more extensive? Or with the cli tests it will be enough?

tests/test_login.py
uriii3: I mean we don't need both but if we overwrite the self.output we destroy the test, no? (or is it appended and then my question makes no sense?)

tests/test_login.py
renaudjester: Basically the CLI and the python interface sends to the same function to I am checking that the arguments work

tests/test_login.py
renaudjester: What do you mean we destroy the test?

renaudjester: And no it's not appended, but in the whole test suit we do: `self.output = `

uriii3 (approved):

PR #163, opened 2024-10-10
test: try to add coverage

renaudjester:
Try to see if coverage works out of the box:

Coverage works but: 

- Seems to slow down the tests (not sure because had some cloudferro issues this day)
- add xdist to accelerate -> to add seems very efficient

To do for coverage:

- [ ] Integrate it in a workflow
- [ ] Get the answer 
- [ ] Print it somewhere

Tried also to add tests on ubuntu and macOS:

works but way slower due to the use of machines in the US (github runners) see https://www.notion.so/lobelians/Email-for-communication-about-v2-1a2e79950d46439ebc780c421270a8ac?pvs=4
Problem: make the tests that have a timeout (on the describe and the `arco-is-fast` to fail) 
Solutions: 

- have a self hosted mac, windows etc (from mercator?)
- have two test commands and change the tests if not on self-hosted runners
     * use env variable? 
     * Can we exclude tests from a pytest command? 

So have one self-hosted test with full test suit and coverage and then 3 others with macOS * 2 and windows

So far with xdist (no coverage):

- self-hosted: 5min ✅ 
- ubuntu: 8min 🔴  (arco-is-fast is failing)
- mac latest:  12min 🔴  (arco-is-fast is failing)
- macos 13: 14min 🔴  (arco-is-fast is failing)


uriii3:
The coverage was implemented here: https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/352 and the different OS is also being implemented in https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/355 . This is covered!

PR #167, opened 2024-10-15
add repository package metadata

jacopo-exact:
add the repository field in the package metadta, so that PyPI will automatically add it to the package page and improve discoverability of the repository.


renaudjester (approved):
Nice :D Thanks for this! I redirected it to the v2 branch! let's release it with the version 2 of the toolbox (since for example the v1 doesn't have documentation)

uriii3:
It doesn't like the credentials use, no?

renaudjester:
We are looking into how to do the test suit for pull requests coming from forks. I'll come back to you.

Worst case I will integrate your changes manually to the repo

PR #172, opened 2024-10-17
test: tests can work from fork

renaudjester:
from this: https://stackoverflow.com/questions/76975408/how-to-pass-a-secret-from-a-forked-github-repo-to-source-repo 

To be able to run the tests for forked branches


uriii3:
[Here](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) (one link in the same issue) i'm not fully understansing how 'vulnerable' the code would be. Looks like it is a bit unsure, no?

PR #175, opened 2024-10-18
fix: small changes docs

uriii3:
I was going through the word again and I found that maybe these things weren't adressed, don't know if it is worth merging some of them or if we consider that it was actually okay as they were!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 ## Features
 
-The `copernicusmarine` Toolbox offers capabilities through both **Command Line Interface (CLI)** and **Python API**:
+The Copernicus Marine toolbox offers capabilities through both **Command Line Interface (CLI)** and **Python API**:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,9 +11,9 @@
 
 ## Features
 
-The `copernicusmarine` offers capabilities through both **Command Line Interface (CLI)** and **Python API**:
+The `copernicusmarine` Toolbox offers capabilities through both **Command Line Interface (CLI)** and **Python API**:
```
renaudjester, line 14:
```suggestion
The Copernicus Marine toolbox offers capabilities through both **Command Line Interface (CLI)** and **Python API**:
```

renaudjester (approved):

README.md
uriii3: Screenshot 2024-10-18 at 17 38 50
I am not sure, but the report they did explicitly wrote it the way I did.

PR #176, opened 2024-10-18
fix: correcting binaries

uriii3:
I think there needs to be more than one version for the ubuntu machines.


uriii3:
In the end, I made the binary for the lowest ubuntu version possible in the GitHub runners. In the commits, there is also the code to create a Docker Image and try to create the binary there... it looks like the Ubuntu machines will be updated to the newer requirements, let's see what we discuss in the meeting.

.github/workflows/binaries.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, macos-13]
+        os: [macos-latest, ubuntu-latest, macos-13, ubuntu-20.04]
```
renaudjester, line 14: You should probably detail the limitation and the details in the documentation

Makefile
```diff
@@ -71,6 +71,12 @@ run-using-pyinstaller-ubuntu-latest:
...
 	chmod +rwx /home/runner/work/copernicus-marine-toolbox/copernicus-marine-toolbox/dist/copernicusmarine_linux.cli
 
+run-using-pyinstaller-ubuntu-20.04:
+	pip install -e .
+	ldd --version
+	python3 -m PyInstaller --collect-all tzdata --copy-metadata copernicusmarine --icon=toolbox_icon.png --name copernicusmarine_linux-20-04.cli --add-data="/home/runner/micromamba/envs/copernicusmarine-binary/lib/python3.12/site-packages/distributed/distributed.yaml:./distributed"  copernicusmarine/command_line_interface/copernicus_marine.py --onefile --path /opt/hostedtoolcache/Python/3.12/x64/lib/python3.12/site-packages --copy-metadata xarray
+	chmod +rwx /home/runner/work/copernicus-marine-toolbox/copernicus-marine-toolbox/dist/copernicusmarine_linux-20-04.cli
+
```
renaudjester, line 79: Same as the ubuntu latest right? Then would be probably better to use to variables to not repeat the code

Makefile
```diff
@@ -71,6 +71,12 @@ run-using-pyinstaller-ubuntu-latest:
 	python3 -m PyInstaller --collect-all tzdata --copy-metadata copernicusmarine --icon=toolbox_icon.png --name copernicusmarine_linux.cli --add-data="/home/runner/micromamba/envs/copernicusmarine-binary/lib/python3.12/site-packages/distributed/distributed.yaml:./distributed"  copernicusmarine/command_line_interface/copernicus_marine.py --onefile --path /opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages --copy-metadata xarray
 	chmod +rwx /home/runner/work/copernicus-marine-toolbox/copernicus-marine-toolbox/dist/copernicusmarine_linux.cli
```
renaudjester, line 72: rename this with something like "ubuntu_22.04" to be consistent with the other one! Or whatever name is relevevant for the user to know which one they can use 

Maybe it's not only about ubuntu and the name "linux" is better, I don't know  but needs to be investigated

Makefile
uriii3: Mmh well, not really because here I'm creating the binary itself and the name it has, is different in both cases. I made a similar approach to the macOS. If for you it is mandatory to have only one command I can battle with it.

renaudjester: That's what I am saying: you can use a variable for the name of the file (similar to what is done with `BUMP_TYPE` for example) 
It shouldn't be too much of a fight and should make it easier to maintain

PR #182, opened 2024-10-22
chore: [get, subset] add skip existing option

uriii3:
Add a skip-existing file option for get and subset.

Solving [CMT-115](https://cms-change.atlassian.net/browse/CMT-115).

- [x] Add option for subset
- [ ] Add option for get
- [ ] Define possible interactions with overwrite-output-data and sync options
- [ ] Add tests to see the behaviour
- [ ] Restructure tests if needed
- [ ] Add documentation


uriii3:
For the subset, define the interaction between `skip_existing` and `overwrite_output_data`:
- Currently, the interaction is defined to prioritise the `overwrite_output_data`. That way, if this option is set to true, the code will overwrite the files that are the same, independently if the option `skip_existing` is set to True or not.
- If `dry_run` is set, no data will be downloaded for sure. Still, it will check if the file exists or not, showing a different message in those cases.
- If no `dry_run` and no `overwrite_output_data` but `skip_existing` is set to True, it will not download the file if it is already in the output directory specified.

I added a file within the SubsetResponse, `download_status`. This will display a different message, defined in `get_message_data_downloaded(...)`, regarding the status of the data downloaded.

PR #184, opened 2024-10-23
feat: support the new authentification system

renaudjester:
[Fix CMT-4](https://cms-change.atlassian.net/browse/CMT-4)

### DIFF
--- a/tests/test_login.py
+++ b/tests/test_login.py
@@ -290,6 +290,34 @@ class TestLogin:
             b"Valid credentials from configuration file" in self.output.stderr
         )
 
+    def test_login_falls_back_to_old_system(self):
+        environment_without_crendentials = (
+            get_environment_without_crendentials()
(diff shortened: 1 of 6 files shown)


tests/test_login.py
```diff
@@ -290,6 +290,34 @@ def check_credentials_file_is_valid(self, tmp_path):
             b"Valid credentials from configuration file" in self.output.stderr
         )
 
+    def test_login_falls_back_to_old_system(self):
```
uriii3, line 293: should we ask for an account here so that we can check that we can connect through the new authentication system?

uriii3 (changes requested):

tests/test_login.py
renaudjester: You account already works

uriii3 (approved):

PR #186, opened 2024-10-24
chore: update the deprecation for v2

renaudjester:
Add deprecation for options that will be deleted in v2. 

Added some different messages depending on the cases: 

- If it has a replacement already
- If it will be deleted or deprecated in the v2
- If it is only about the v2

### DIFF
--- a/copernicusmarine/core_functions/deprecated.py
+++ b/copernicusmarine/core_functions/deprecated.py
@@ -31,6 +31,8 @@ def get_deprecated_message(
         message += (
             "This option will no longer be "
             + "available in copernicusmarine>=2.0.0. "
+            + "Please refer to the documentation for more information "
+            + "when the new major version is released."
         )
     if deprecated_for_v2:
         message += (
--- a/copernicusmarine/core_functions/deprecated.py
+++ b/copernicusmarine/core_functions/deprecated.py
@@ -31,8 +31,8 @@ def get_deprecated_message(
         message += (
             "This option will no longer be "
             + "available in copernicusmarine>=2.0.0. "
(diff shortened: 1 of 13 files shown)


copernicusmarine/core_functions/deprecated.py
```diff
@@ -40,26 +80,43 @@ def make_parser(self, ctx):
...
         options |= set(parser._long_opt.values())
 
+        # get name of the command
+        command_name = ctx.command.name
+
         for option in options:
-            if not isinstance(option.obj, DeprecatedClickOption):
-                continue
+            # if not isinstance(option.obj, DeprecatedClickOption):
+            #     continue
```
uriii3, line 90: we can delete, no?

copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -69,5 +92,70 @@ def dict_old_names_to_new_names(self):
...
+            old_name="netcdf_compression_enabled",
+            new_name="netcdf_compression_enabled",
+            only_for_v2=True,
+        ),
+        DeprecatedOption(
+            old_name="disable_progress_bar",
+            new_name="disable_progress_bar",
+            targeted_functions=["open_dataset", "read_dataframe"],
+            only_for_v2=True,
+        ),  # TOCHECK: shouldn't be triggered for all
```
uriii3, line 131: make sure to check this too

uriii3 (changes requested):

uriii3:
Looks good to me!

uriii3 (approved):

PR #195, opened 2024-10-28
chore: adding more binaries

uriii3:
I think there needs to be more than one version for the ubuntu machines.

In the end, I made the binary for the lowest ubuntu version possible in the GitHub runners. In the commits, there is also the code to create a Docker Image and try to create the binary there... it looks like the Ubuntu machines will be updated to the newer requirements, let's see what we discuss in the meeting.


uriii3:
Hehe I commented your posts on the other branch xd. I was saying, about having two distinct commands in the make file:

Mmh well, not really because here I'm creating the binary itself and the name it has, is different in both cases. I made a similar approach to the macOS. If for you it is mandatory to have only one command I can battle with it.

uriii3:
I merged both the macOS and the linux orders into one, I put the correct name (Linux for now as we know and as reported in the PyInstaller page) and I corrected the way to exclude the branches so that it will be handled correctly.

doc/installation.rst
```diff
@@ -119,3 +119,11 @@ And from a Windows os (cmd):
...
 
 (``describe`` or any other command)
+
+.. note::
+
+    For the **Linux** binaries you need a version of Glibc higher than the one that the binary has. Use older **Linux** versions to access a lower version of said library.
+
+.. note::
+
+    The **Linux** binaries are tested with Ubuntu machines.
```
renaudjester, line 129: There is not a way to know the Glibc version of our binaries? 

Also, maybe you can add some info on how people can check the version they have

Makefile
```diff
@@ -54,6 +54,31 @@ run-tests-dependencie-versions:
...
+
+run-using-pyinstaller-linux:
+	pip install -e .
+	python3 -m PyInstaller --collect-all tzdata --copy-metadata copernicusmarine --icon=toolbox_icon.png --name copernicusmarine_${DISTRIBUTION}.cli --add-data="/home/runner/micromamba/envs/copernicusmarine-binary/lib/python3.12/site-packages/distributed/distributed.yaml:./distributed"  copernicusmarine/command_line_interface/copernicus_marine.py --onefile --path /opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages --copy-metadata xarray
+	chmod +rwx /home/runner/work/copernicus-marine-toolbox/copernicus-marine-toolbox/dist/copernicusmarine_${DISTRIBUTION}.cli
+
+run-using-pyinstaller-ubuntu-22.04: DISTRIBUTION = linux-22.04
+run-using-pyinstaller-ubuntu-22.04: run-using-pyinstaller-linux
+
+run-using-pyinstaller-ubuntu-20.04: DISTRIBUTION = linux-20.04
```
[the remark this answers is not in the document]
renaudjester: I don't really know how to solve it but yeah `22.04` is specific to ubuntu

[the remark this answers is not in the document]
uriii3: 😞 😞 

Mmh what do you think `linux-newer` and `linux-older` ?  to differentiate both of them

doc/installation.rst
uriii3: Mmh yes, I can print it within the creation of the binaries and print it later... and also point them the way to do it.

uriii3: I thought that maybe it was too technical and that they didn't need to understand the logic of it behind it, but makes sense to pin point the people in the right direction.

renaudjester: Well to me it makes sense to say: 
- here are the version we have: x.y.z so we support >=x.y.z
- here is how you know yours: `some comand`

Doesn't need to go into details more than this but at least it's easy to check

uriii3:
I hardcoded the version of the GLIBC library into the code. We can upgrade the names of the binaries from the release versions and it will always print it so we could check when we release. 

If we see that it changes a lot we can also investigate how to automate it.

Makefile
```diff
@@ -54,6 +54,32 @@ run-tests-dependencie-versions:
...
+	pip install -e .
+	ldd --version
+	python3 -m PyInstaller --collect-all tzdata --copy-metadata copernicusmarine --icon=toolbox_icon.png --name copernicusmarine_${DISTRIBUTION}.cli --add-data="/home/runner/micromamba/envs/copernicusmarine-binary/lib/python3.12/site-packages/distributed/distributed.yaml:./distributed"  copernicusmarine/command_line_interface/copernicus_marine.py --onefile --path /opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages --copy-metadata xarray
+	chmod +rwx /home/runner/work/copernicus-marine-toolbox/copernicus-marine-toolbox/dist/copernicusmarine_${DISTRIBUTION}.cli
+
+run-using-pyinstaller-ubuntu-22.04: DISTRIBUTION = linux-glibc-2.35
+run-using-pyinstaller-ubuntu-22.04: run-using-pyinstaller-linux
+
+run-using-pyinstaller-ubuntu-20.04: DISTRIBUTION = linux-glibc-2.31
+run-using-pyinstaller-ubuntu-20.04: run-using-pyinstaller-linux
```
renaudjester, line 81: I would avoid "." in the file name (except for the extension obviously)

https://stackoverflow.com/questions/3926324/dot-in-file-name

renaudjester (approved):

PR #198, opened 2024-10-29
feat: upgrade the dry run and returned fields

renaudjester:
Fix [CMT-105](https://cms-change.atlassian.net/browse/CMT-105)


uriii3 (approved):
Niceeee (have you also checked if there is any way to correct the format of the json file when doing the `dry-run`?)

PR #200, opened 2024-10-30
fix: subset was not using staging metadata

renaudjester:
Even with `copernicusmarine subset --staging` the command was not getting the dataset from staging

Bug introduced when got rid of the cache

### DIFF
--- a/tests/test_command_line_interface.py
+++ b/tests/test_command_line_interface.py
@@ -1860,3 +1860,37 @@ class TestCommandLineInterface:
         assert datetime.datetime.strptime(
             str(dataset.time.values.max()), "%Y-%m-%dT%H:%M:%S.000%f"
         ) >= datetime.datetime.strptime("2023-01-03", "%Y-%m-%d")
+
+    def test_get_goes_to_staging(self):
+        command = [
+            "copernicusmarine",
+            "get",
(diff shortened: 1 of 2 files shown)


tests/test_command_line_interface.py
```diff
@@ -1860,3 +1860,37 @@ def test_coordinates_selection_method_outside_w_elevation(self, tmp_path):
...
+            "cmems_mod_ibi_phy_my_0.083deg-3D_P1Y-m",
+            "--staging",
+            "--log-level",
+            "DEBUG",
+        ]
+        self.output = execute_in_terminal(command)
+        assert (
+            b"mdl-metadata-dta/dataset_product_id_mapping.json"
+            in self.output.stderr
+        )
```
uriii3, line 1878: doesn't it change sometimes, the staging? I guess it won't be a problem though, no? It will be like when there is a change in the metadata right now

uriii3 (approved):

tests/test_command_line_interface.py
renaudjester: This specific file won't change

renaudjester:
Bug introduced when got rid of the cache: no need to explain in the change log

PR #201, opened 2024-10-30
fix: use coordinate equivalent

renaudjester:
I am very surprise this hasn't been reported before, I guess a vast majority of the datasets follow the conventional names

### DIFF
--- a/copernicusmarine/download_functions/subset_xarray.py
+++ b/copernicusmarine/download_functions/subset_xarray.py
@@ -200,12 +200,12 @@ def get_size_of_coordinate_subset(
     minimum: Optional[Union[float, DateTime]],
     maximum: Optional[Union[float, DateTime]],
 ) -> int:
-    for label in COORDINATES_LABEL[coordinate]:
-        if label in dataset.sizes:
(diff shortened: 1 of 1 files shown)


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -200,12 +200,12 @@ def get_size_of_coordinate_subset(
...
             return (
-                dataset.coords[coordinate]
-                .sel({coordinate: slice(minimum, maximum)}, method=None)
-                .coords[coordinate]
+                dataset.coords[coordinate_label]
+                .sel({coordinate_label: slice(minimum, maximum)}, method=None)
+                .coords[coordinate_label]
                 .size
             )
     else:
```
uriii3, line 211: does this actually work? a `for: ... else: ...` ? I will check now but maybe the problem is that the error doesn't raise anyway. Let me check

uriii3 (approved):

uriii3 (commented):
What sense does it have to have a function that is being used in `utils.py` defined in `subset_xarray.py`?

uriii3 (approved):

PR #205, opened 2024-11-07
chore: delete force download and show ouputnames options

renaudjester:
fix [CMT-131](https://cms-change.atlassian.net/browse/CMT-131)


copernicusmarine/download_functions/download_arco_series.py
```diff
@@ -117,17 +114,11 @@ def download_dataset(
...
-        logger.info(message_formatted_dataset_size_estimation)
-        click.confirm(
-            FORCE_DOWNLOAD_CLI_PROMPT_MESSAGE,
-            default=True,
-            abort=True,
-            err=True,
-        )
     else:
-        logger.info(message_formatted_dataset_size_estimation)
+        logger.debug(dataset)
```
uriii3, line 120: why this?

renaudjester: Because before we were showing more information with the prompt: 

if you were doing subset without force download you would see the dataset and the prompt
if you were doing subset with force download you don't see the dataset nor the prompt. Now the user can see it if they use debug log level or using dry-run in info log level

PR #209, opened 2024-11-12
chore: add binary check

uriii3:
Adding binaries checks, first for macOS and ubuntu, while still making them work


.github/workflows/binary-test-windows.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+      - "release/**"
+      - "pre-releases/**"
+  pull_request: # for the moment, to test that it works.
+    branches:
+      - "main"
```
renaudjester, line 12: Check also if the `pyproject.toml` has been modified, if so, trigger

tests_binaries/test_windows_binaries.py
renaudjester, line 1: I don't think you need different ones do you?

.github/workflows/binary-test-windows.yml
uriii3: I assumed that if `pyproject.toml` changed was because a push in main, so I decided to skip it. Do you think it is good enough?

tests_binaries/test_windows_binaries.py
uriii3: Mmh well, no, but yes. So when doing the checks, the terminal command is: `./copernicusmarine.cli` or `./copernicusmarine.exe`, in the macOS and ubuntu first and then in windows. There is a way to pass parameters to the purest but I think it complicated more everything than made things easier, because  you needed a `conffile.py` or something like that and I was scared to change the purest for all the repo.

renaudjester: You could pass an environment variable with the name of the file and then you open it in the tests :D

uriii3: This [stack-overflow](https://stackoverflow.com/questions/36141024/how-to-pass-environment-variables-to-pytest) gives the impression that it is a little bit more complex... looks weird that the easy answer (third one) is not the most upvoted, but I'll try anyway.

.github/workflows/binary-test-windows.yml
```diff
@@ -0,0 +1,48 @@
+name: Run tests binaries windows
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+      - "release/**"
+      - "pre-releases/**"
+  pull_request:
```
renaudjester, line 10: you need a `pull_request_target` trigger type otherwise your tests will not work on a fork (you can check the `tests.yml` to see what you need to add

renaudjester (approved):

PR #214, opened 2024-11-15
chore: reordering options

uriii3:
Reordering the options in which the parameters appear in the documentation. So:
subset.cli options are ordered the same as subset.api.

I didn't order equally the cli page with the api page because I prefer the order in the cli but it is hard to change the order of the API, and maybe not the highest priority to invest my time in this. Will rest as a TODO.


renaudjester (approved):
You can merge this is you want but I think that we will need a full review of the order and organisation of the options

uriii3:
I think it actually makes a little sense how they are ordered, I checked a bit and it is not bad... so I think I will merge this, in case we just want to keep it this way!

PR #215, opened 2024-11-15
fix: netcdf compression solved

uriii3:
With the help of xarray developers in [Issue #9783](https://github.com/pydata/xarray/issues/9783) we solved the issue with the encoding for netcdf datasets (it looked like it was already pretty well compressed and we were not sending all the information to the output).

I also changed the default value to 4 for the CLI subset because from [here](https://unidata.github.io/netcdf4-python/#netCDF4.Dataset.createVariable) it looked like the default compression value for xarray is also 4.

Looks like now, indeed, the 3 commands are saving the files with 'according' sizes:
```
copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m -v thetao -t 1993-01-01T00:00:00 -T 2020-12-31T23:59:59 -x -90 -X -85 -y -35 -Y -30 -z 0.49 -Z 1 --netcdf-compression-level 1 -f comlevel1.nc

copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m -v thetao -t 1993-01-01T00:00:00 -T 2020-12-31T23:59:59 -x -90 -X -85 -y -35 -Y -30 -z 0.49 -Z 1 -f complevel0.nc

copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m -v thetao -t 1993-01-01T00:00:00 -T 2020-12-31T23:59:59 -x -90 -X -85 -y -35 -Y -30 -z 0.49 -Z 1 --netcdf3-compatible -f netcdf3.nc
```

### DIFF
--- a/copernicusmarine/download_functions/download_arco_series.py
+++ b/copernicusmarine/download_functions/download_arco_series.py
@@ -559,8 +559,8 @@ def _download_dataset_as_netcdf(
         encoding = {
             name: {
                 **{
-                    key: var.encoding[key]
-                    for key in var.encoding.keys()
+                    key: value
+                    for key, value in var.encoding.items()
                     if key in keys_to_keep
                 },
                 **comp,
--- a/copernicusmarine/download_functions/download_arco_series.py
+++ b/copernicusmarine/download_functions/download_arco_series.py
@@ -495,7 +495,6 @@ def _save_dataset_locally(
     netcdf_compression_level: int,
     netcdf3_compatible: bool,
 ):
-    logger.info(output_path)
     if output_path.suffix == ".zarr":
         if netcdf_compression_level > 0:
             raise NetCDFCompressionNotAvailable(
--- a/copernicusmarine/download_functions/download_arco_series.py
+++ b/copernicusmarine/download_functions/download_arco_series.py
@@ -539,22 +539,30 @@ def _download_dataset_as_netcdf(
             contiguous=False,
             shuffle=True,
         )
-        encoding = {
+        long_encoding = {
             name: {**var.encoding, **comp}
             for name, var in dataset.data_vars.items()
         }
         for var in dataset.data_vars:
-            for key in [
-                "szip",
-                "zstd",
-                "bzip2",
-                "blosc",
-                "preferred_chunks",
-                "compressor",
-                "filters",
-                "coordinates",
-            ]:
-                encoding[var].pop(key, None)
+            keys_to_keep = [
+                "dtype",
+                "blosc_shuffle",
+                "shuffle",
+                "chunksizes",
+                "szip_coding",
+                "zlib",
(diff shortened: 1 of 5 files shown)


uriii3:
I'm scared that the dictionary of 'out' keys might not be extensive enough, should it be better to do a dictionary with THE keys that can go in? Can then be that we run into the same problem -> we are not extensive enough? But then it wouldn't ERROR so it would not be that bad.

uriii3:
I finally opted to have a positive definition of the keys, looks more robust!

copernicusmarine/download_functions/download_arco_series.py
```diff
@@ -538,7 +538,35 @@ def _download_dataset_as_netcdf(
...
+                "compression_opts",
+                "compression",
+                "quantize_mode",
+                "endian",
+            ]
+            encoding[var] = {
+                key: long_encoding[var][key]
+                for key in long_encoding[var].keys()
+                if key in keys_to_keep
+            }
```
renaudjester, line 569:
```suggestion
        keys_to_keep = {
    "_FillValue",
    "blosc_shuffle",
    "chunksizes",
    "complevel",
    "compression",
    "compression_opts",
    "contiguous",
    "dtype",
    "endian",
    "fletcher32",
    "quantize_mode",
    "shuffle",
    "significant_digits",
    "szip_coding",
    "szip_pixels_per_block",
    "zlib",
}
        encoding = {
            name: {**{key: value for key, value in var.encoding if key in keys_to_keep}, **comp}
            for name, var in dataset.data_vars.items()
        }
```

To be formatted and check if it works but the idea is that this part of the code can be more concise. (also notice the user of a set and not a list for the `keys_to_keep`

tests/test_command_line_interface.py
renaudjester, line 1: Maybe you have some tests in mind to check that the size of the file is as expected

copernicusmarine/download_functions/download_arco_series.py
uriii3: I feel like this is complicating too much the code (and to be honest it will be hard for me to actually make it work). Do you think it is necessary to have the code as concise as possible? (even if it get's more tricky to maintain and develop 🌚 ?)

tests/test_command_line_interface.py
uriii3: Mmh yes, there actually existed checks but maybe I can add this new ones (which have some extra keywords, so the datasets already come with some compression, that we should keep). So yes, I will add them!

copernicusmarine/download_functions/download_arco_series.py
uriii3: Mmh okay 🫡

renaudjester (approved):

PR #216, opened 2024-11-18
chore: improving the catch of the error

uriii3:
I tried to improve the catch on the error of Xarray. Instead of catching it, though, I assert before setting the corresponding engine that such library exists.

Pros:
- assertion seems more manageable than a try-except
- It is more personalised
- The code looks well and understandable

Cons: 
- It is not embedded into xarray but rather we should be the ones updating it if there is any change in the future.

I don't know if you will like it, tell me if you prefer a try-except with the customised exception of xarray!


copernicusmarine/download_functions/download_arco_series.py
```diff
@@ -543,7 +544,13 @@ def _download_dataset_as_netcdf(
...
 
     xarray_download_format = "NETCDF3_CLASSIC" if netcdf3_compatible else None
-    engine = "h5netcdf" if not netcdf3_compatible else "netcdf4"
+    engine = "h5netcdf"
+    if netcdf3_compatible:
+        assert importlib.util.find_spec("netCDF4"), (
+            "To use the NETCDF3_COMPATIBLE option you need to have 'netCDF4' "
+            "installed. You can install it with 'pip install netcdf4'."
+        )
+        engine = "netcdf4"
```
renaudjester, line 553: I would do this at the very beginning of the subset function in `subset.py`

renaudjester (approved):

PR #217, opened 2024-11-19
warn: preventing subset on stereographic data

uriii3:
Stereographic data will soon be added to the MDS backend. The toolbox will develop the feature to be able to subset such datasets, but while we are working on it, it should pop out some sort of warning to the users.

With this PR, it will show a warning when subsetting:
```
WARNING - 2024-11-19T11:53:36Z - Dataset part is in stereographic projection.
```

And it will throw an error if there is some sort of geospatial subsetting:
```
ERROR - 2024-11-19T11:53:36Z - Geospatial subset not available for stereographic projection: The geospatial subset of datasets with stereographic projection is not yet available. We are developing such feature and will be supported in future versions.
```

Looking forward to implement it!

----
📚 Documentation preview 📚: https://copernicusmarine--217.org.readthedocs.build/en/217/


uriii3:
@renaudjester I think we can add some tests when the mapping works as expected. Until then, maybe there is no need to check for the staging... but however you consider it.

copernicusmarine/core_functions/exceptions.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, message: str):
         self.__setattr__("custom_exception_message", message)
 
 
+class GeospatialSubsetNotAvailableForOriginalProjection(Exception):
```
renaudjester, line 44: Add it to the init

renaudjester: So it can be visible in the documentation

copernicusmarine/core_functions/services_utils.py
```diff
@@ -263,6 +263,7 @@ class RetrievalService:
     uri: str
     dataset_valid_start_date: Optional[Union[str, int, float]]
     service: CopernicusMarineService
+    isOriginalGrid: bool = False
```
renaudjester, line 266: should be snake case :D

renaudjester (commented):
Add also two tests at least: one that you can subset stereographic data (on time and depth) and that the error is returned when doing the subset on geographical domain.

copernicusmarine/core_functions/exceptions.py
uriii3: Mmh didn't we say that stereographic was too narrow, so we would maintain original projection as a broader term that will include all the options?

copernicusmarine/core_functions/exceptions.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, message: str):
...
+    The part you want to subset has a original projection.
+
+    The geospatial subset of such datasets is not yet available.
+    """
+
+    def __init__(self):
+        super().__init__(
+            "The geospatial subset of datasets with original projection is not "
+            "yet available. "
+            "We are developing such feature and will be supported in future versions."
```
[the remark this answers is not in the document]
uriii3: done manually hehe

copernicusmarine/core_functions/exceptions.py
```diff
@@ -41,6 +41,21 @@ def __init__(self, message: str):
         self.__setattr__("custom_exception_message", message)
 
 
+class GeospatialSubsetNotAvailableForNonLatLon(Exception):
+    """
+    The part you want to subset has a original projection.
```
renaudjester, line 46: What is an original projection? 

"The data you are requesting is using a projection that is not on the normalised latitude and longitude grid. The geospatial subset of such datasets is not yet available. 

Please check other parts of the dataset to subset it. The geospatial subset of the datasets with different gridding will be fully available soon. "

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -597,6 +599,16 @@ def check_dataset_subset_bounds(
         dataset_coordinates = dataset.coords
     else:
         raise ServiceNotSupported(service_name)
+    if is_original_grid:
+        logger.warning("Dataset part has the original projection.")
```
renaudjester, line 603: debug maybe better?

tests/test_originalGrid_datasets.py
```diff
@@ -0,0 +1,99 @@
...
+
+import pendulum
+import xarray
+
+from copernicusmarine.core_functions.utils import datetime_parser
+from tests.test_utils import execute_in_terminal
+
+
+class TestOriginalGridDatasets:
+    def test_cmt_identifies_originalGrid_datasets(self):
```
renaudjester, line 12:
```suggestion
    def test_toolbox_identifies_originalGrid_datasets(self):
```

tests/test_originalGrid_datasets.py
```diff
@@ -0,0 +1,99 @@
...
+        assert (
+            b"Dataset part has the original projection" in self.output.stderr
+        )
+        assert b"ERROR" in self.output.stderr
+        assert (
+            b"Geospatial subset not available for non lat lon: The "
+            b"geospatial subset of datasets in a projection that is not in "
+            b"latitude and longitude is not yet available. We are "
+            b"developing such feature and will be supported in future versions."
+        ) in self.output.stderr
```
renaudjester, line 61: What about the geographical extend in the response of the subset? maybe it will be latitude and longitude no?

tests/test_originalGrid_datasets.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_originalGrid_error_when_timepart(self, tmp_path):
```
renaudjester, line 63:
```suggestion
    def test_originalGrid_works_when_time_and_depth_subsetting(self, tmp_path):
```

tests/test_originalGrid_datasets.py
uriii3: meh well in this case (it errors) so it will not have any geographical extend. In the other one, I check that the size is greater than ', but I didn't know what else to check.

copernicusmarine/core_functions/models.py
```diff
@@ -156,6 +156,7 @@ class GeographicalExtent(BaseModel):
     minimum: Optional[float]
     maximum: Optional[float]
     unit: Optional[str]
+    id: Optional[str]
```
renaudjester, line 159: better something more explicit like `coordinate_id`(like we said: use the same name as we use in the describe types)

copernicusmarine/download_functions/utils.py
```diff
@@ -225,12 +233,15 @@ def _get_coordinate_extent(
...
                     unit=unit,
+                    id=coord_label,
                 )
             return GeographicalExtent(
                 minimum=minimum,
                 maximum=maximum,
                 unit=unit,
+                id=coord_label,
             )
+    logger.info("passant per aquí")
```
renaudjester, line 244: comment

copernicusmarine/download_functions/utils.py
```diff
@@ -190,17 +189,26 @@ def _format_datetimes(
...
-        time=_get_coordinate_extent(dataset, "time"),  # type: ignore
-    )
-    depth_or_elevation_extent = _get_coordinate_extent(dataset, "depth")
-    if "depth" in dataset.sizes:
-        coordinates_extent.depth = depth_or_elevation_extent  # type: ignore
-    elif "elevation" in dataset.sizes:
-        coordinates_extent.elevation = depth_or_elevation_extent  # type: ignore
+) -> list[Union[GeographicalExtent, TimeExtent, None]]:
+    coordinates_extent = []
+    for coord_label in dataset.sizes:
```
renaudjester, line 194: Two things: 

- I don't see why you need this for loop and all those "ifs" because in the next function (`_get_coordinate_extent`) you also have a loop and "ifs"

- In general I think we want to avoid looping over the dataset.sizes (it's quite error prone because it depends on the data) whereas here we no that the dataset will have a maximum of four dimensions that are x,y,z,t

uriii3: I tried to solve it iterating through a list of `["longitude", "latitude", "depth", "time"]`.  This way, I check all the dimensions and add them if needed (it was throwing a weird error but I solved it hehe).

tests/test_originalGrid_datasets.py
renaudjester, line 1: Add a test that you have "x" and "y" and not latitude longitude in the result and it should be good!

tests/test_originalGrid_datasets.py
```diff
@@ -0,0 +1,108 @@
...
+        assert self.output.returncode == 0
+        assert b"DEBUG" in self.output.stderr
+        assert (
+            b"Dataset part has the non lat lon projection."
+            in self.output.stderr
+        )
+        assert pendulum.parse("2022-01-01") == datetime_parser(
+            dataset.time.values[0]
+        )
+        assert len(dataset.x.values) > 0
```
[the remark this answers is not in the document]
renaudjester: Maybe check that when using the CLI, when you do a subset the object that is returned only have "x" and "y" cooordinate extent and not latitude longitude one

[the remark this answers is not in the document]
renaudjester: What you are testing now with this line is if the dataset you saved has "x" and "y" coordinate, I meant testing the response of the subset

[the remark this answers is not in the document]
uriii3: Well, actually it doesn't. But I checked doing and open_dataset (with xarray) and the result was the same... so we will need to look at it, but for now seems like the behaviour is somehow standard.

So, the dataset returned does have longitude and latitude but they are 'empty'. Still existing though.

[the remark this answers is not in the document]
renaudjester: Can you test the response of the subset? that you have all the coordinateExtent objects you need and not more?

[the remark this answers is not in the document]
uriii3: Aaah okay, still not understanding the response as the response when you refer to it sorry.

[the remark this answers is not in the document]
renaudjester: My fault I said "result" at first :D

copernicusmarine/core_functions/models.py
```diff
@@ -156,6 +156,7 @@ class GeographicalExtent(BaseModel):
     minimum: Optional[float]
     maximum: Optional[float]
     unit: Optional[str]
+    coordinate_id: Optional[str]
```
renaudjester, line 159: Maybe this is not optional 🤔

copernicusmarine/core_functions/models.py
```diff
@@ -200,7 +187,9 @@ class ResponseSubset(BaseModel):
     #: Variables of the subsetted dataset.
     variables: list[str]
     #: The bounds of the subsetted dataset.
-    coordinates_extent: DatasetCoordinatesExtent
+    coordinates_extent: Optional[
+        list[Union[GeographicalExtent, TimeExtent, None]]
+    ]
```
renaudjester, line 192: not sure this should be optional 🤔

copernicusmarine/catalogue_parser/fields_query_builder.py
```diff
@@ -55,29 +63,50 @@ def build_query(
...
+    ) -> Optional[dict[Type, Literal["__all__", None]]]:
+        models = {}
+        if check_type_is_base_model(type_to_check):
+            return {type_to_check: "__all__" if in_an_iterable else None}
+        elif get_origin(type_to_check) is list:
+            result = self._get_base_models_in_type(
+                get_args(type_to_check)[0], in_an_iterable=True
+            )
+            if result:
+                models.update(result)
```
uriii3, line 98: I didn't know you could update a dictionary like this, looks pretty cool to me!

copernicusmarine/catalogue_parser/fields_query_builder.py
```diff
@@ -55,29 +63,50 @@ def build_query(
...
+                    )
                 result = self.build_query(
-                    get_args(field_type)[0],
-                    query[field_name]["__all__"],
+                    base_model,
+                    (
+                        query[field_name]["__all__"]
+                        if in_an_iterable
+                        else query[field_name]
+                    ),
```
uriii3, line 79: I gotta say that I'm quite relieved that you finished the work, it looks quite complex to me even looking at the code. The query and the builder in itself are things that are hard to think about in a non literal way.

PR #219, opened 2024-11-19
fix: updating link for the binaries

uriii3:
Updating the binaries link in the documentation for the following release.

I'm not sure I like it fully because the link will not work properly for future pre-releases (that is also why I hardcoded them before). I will check it in the future.

----
📚 Documentation preview 📚: https://copernicusmarine--219.org.readthedocs.build/en/219/


uriii3:
This PR is only useful if the next release is a release (and not pre-release!).

renaudjester (approved):

PR #221, opened 2024-11-19
test: fix tox tests

renaudjester:
Tox tests fix: they are used to check that the toolbox's basic commands work for different packages versions and python versions. 
fix [CMT-89](https://cms-change.atlassian.net/browse/CMT-89)

----
📚 Documentation preview 📚: https://copernicusmarine--221.org.readthedocs.build/en/221/


.github/workflows/dependencies-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: Dependencies tests
+
+on:
+  schedule:
+    - cron: '0 0 * * 6'
+
+jobs:
+  test-dependencies-combinations:
+    runs-on: ubuntu-latest
```
uriii3, line 9: do we want to use this as a matrix too? would be a great way to check on other environments, no?

copernicusmarine/download_functions/download_original_files.py
```diff
@@ -268,7 +268,7 @@ def download_files(
         run_concurrently(
             _download_one_file,
             [
-                (username, endpoint_url, bucket, in_file, out_file)
+                (username, endpoint_url, bucket, in_file, str(out_file))
```
uriii3, line 271: isn't it a little bit more error prone if instead of passing a path lib.Path we pass a string? or in some sort of version this means an error?

noxfile.py
```diff
@@ -0,0 +1,51 @@
+import nox
+
+PYTHON_VERSIONS = ["3.9", "3.10", "3.11", "3.12", "3.13"]
+XARRAY_VERSIONS = ["2023.4.0", "latest"]
+DASK_VERSIONS = ["2022.1.0", "latest"]
+BOTO3_VERSIONS = ["1.26.0", "latest"]
+NUMPY_VERSIONS = ["1.26", ">=2.0.0"]
```
uriii3, line 7: why this 4 dependencies and None more? distributed, pedantic, pendulum, Zarr, h5netcdf... what makes a library worth checking?

uriii3:
In general I approve though! hehehe

uriii3 (approved):

.github/workflows/dependencies-tests.yml
renaudjester: Yes it's in my head too but I would say, step by step

copernicusmarine/download_functions/download_original_files.py
renaudjester: Yes it's an error for boto3==1.26

noxfile.py
renaudjester: We can think of more if we think that they are very error prone! right now it's the main ones that are important for the processes

PR #224, opened 2024-11-22
feat: add test script for all the datasets in the catalogue

renaudjester:
Script that runs a subset for all the datasets available. 

fix [CMT-1](https://cms-change.atlassian.net/browse/CMT-1)

----
📚 Documentation preview 📚: https://copernicusmarine--224.org.readthedocs.build/en/224/


tests/tests_on_all_datasets/test_on_all_datasets.py
```diff
@@ -0,0 +1,557 @@
+# import asyncio
```
uriii3, line 1: are we using asyncio again? hehehe

renaudjester: It was for testing but finally no

PR #225, opened 2024-11-22
fix: sync available for datasets with parts

renaudjester:
fix [CMT-151](https://cms-change.atlassian.net/browse/CMT-151) by letting users use the `--sync` option with dataset with parts

### DIFF
--- a/copernicusmarine/core_functions/services_utils.py
+++ b/copernicusmarine/core_functions/services_utils.py
@@ -5,6 +5,7 @@ from itertools import chain
 from typing import List, Literal, Optional, Tuple, Union
 
 from copernicusmarine.catalogue_parser.catalogue_parser import (
+    PART_DEFAULT,
     CopernicusMarineCatalogue,
     CopernicusMarineDatasetServiceType,
     CopernicusMarineDatasetVersion,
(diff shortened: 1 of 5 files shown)


copernicusmarine/core_functions/services_utils.py
```diff
@@ -436,25 +431,22 @@ def _get_retrieval_service_from_dataset_version(
...
     if (
-        force_service_type == CopernicusMarineDatasetServiceType.FILES
+        len(dataset_version.parts) > 1
+        and dataset_sync
         and not force_dataset_part_label
-        and not index_parts
-        and len(dataset_version.parts) > 1
+        and not [
+            part for part in dataset_version.parts if part.name == PART_DEFAULT
+        ]
```
uriii3, line 444: So, if we have more than one part and we are syncing, they should either specify the part they one or there should be a default part, no?

My point is, the last one should be an 'or'? So, either there is a default part or they are forcing the part. But no need to have them both at the same time, no?

uriii3 (approved):

copernicusmarine/core_functions/services_utils.py
renaudjester: Not sure because what the if does raise so in other words: 
if the dataset have more than one part
and we are in sync mode
and there is no default part
and we haven't asked for part
=> won't work

If any of this is true, it works: if I have a default part or I asked for a specific one it will work or if am not in sync mode or if there is only one part

PR #226, opened 2024-11-22
chore: add deprecated options to prepare for v2

renaudjester:
Add deprecation warnings for deleted options in v2: [CMT-152](https://cms-change.atlassian.net/browse/CMT-152)

### DIFF
--- a/copernicusmarine/core_functions/deprecated_options.py
+++ b/copernicusmarine/core_functions/deprecated_options.py
@@ -8,7 +8,6 @@ class DeprecatedOption:
         old_name,
         new_name,
         replace=True,
-        deprecated_for_v2=False,
         deleted_for_v2=True,
         only_for_v2=False,
         targeted_functions: Optional[list[str]] = None,
@@ -16,7 +15,6 @@ class DeprecatedOption:
         self.old_name = old_name
         self.new_name = new_name
         self.replace = replace
-        self.deprecated_for_v2 = deprecated_for_v2
         self.deleted_for_v2 = deleted_for_v2
         self.only_for_v2 = only_for_v2
         if not targeted_functions:
@@ -84,14 +82,6 @@ DEPRECATED_OPTIONS: DeprecatedOptionMapping = DeprecatedOptionMapping(
         ),
         DeprecatedOption(old_name="force_service", new_name="service"),
         DeprecatedOption(
-            old_name="service",
-            new_name="service",
-            replace=False,
-            deleted_for_v2=True,
-            only_for_v2=True,
-            targeted_functions=["get"],
-        ),
-        DeprecatedOption(
             old_name="download_file_list",
             new_name="create_file_list",
             replace=False,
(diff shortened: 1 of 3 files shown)


copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -140,30 +138,63 @@ def dict_old_names_to_new_names(self):
...
             new_name="include_all",
-            deprecated_for_v2=True,
-            deleted_for_v2=False,
             only_for_v2=True,
         ),
+        DeprecatedOption(
+            old_name="include_versions",
+            new_name="show_all_versions",
+            only_for_v2=True,
+            replace=False,
```
uriii3, line 162: replace = True?

copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -140,30 +138,63 @@ def dict_old_names_to_new_names(self):
...
+        ),
+        DeprecatedOption(
+            old_name="vertical_dimension_as_originally_produced",
+            new_name="vertical_axis",
+            only_for_v2=True,
+            replace=False,
```
uriii3, line 168: replace = True?

copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -140,30 +138,63 @@ def dict_old_names_to_new_names(self):
...
+        ),
+        DeprecatedOption(
+            old_name="overwrite_output_data",
+            new_name="overwrite",
+            only_for_v2=True,
+            replace=False,
```
uriii3, line 174: replace = True?

copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -140,30 +138,63 @@ def dict_old_names_to_new_names(self):
...
+        DeprecatedOption(
+            old_name="show_outputnames",
+            new_name="show_outputnames",
+            only_for_v2=True,
+        ),
+        DeprecatedOption(
+            old_name="overwrite_configuration_file",
+            new_name="force_overwrite",
+            only_for_v2=True,
+            replace=False,
```
uriii3, line 190: replace = True?

uriii3: it seems to do the same output with True or False, which is sort of weird, no? And it just says that the option will be deprecated, not showing the next one

copernicusmarine/core_functions/deprecated_options.py
```diff
@@ -140,30 +138,63 @@ def dict_old_names_to_new_names(self):
...
+        ),
+        DeprecatedOption(
+            old_name="overwrite",
+            new_name="force_overwrite",
+            only_for_v2=True,
+            replace=False,
```
uriii3, line 196: replace = True?

uriii3 (commented):
It looks like the 'replace=False' only works to change the input option -> wouldn't it be best to include some sort of future change, like: `include-versions` will be deprecated but substituted for a new option `show-all-versions`, with the same behaviour. (or something like this?).

uriii3 (approved):

PR #228, opened 2024-11-25
feat: add tests for query builder

uriii3:
Adding some tests to check that the query builder works as expected.

----
📚 Documentation preview 📚: https://copernicusmarine--228.org.readthedocs.build/en/228/


tests/test_query_builder.py
```diff
@@ -0,0 +1,203 @@
...
+            if function == GET_SET:
+                model = ResponseGet
+            else:
+                model = ResponseSubset
+        queryable_fields = get_queryable_requested_fields(
+            included, model, fields
+        )
+
+        check_intersection = queryable_fields.intersection(function)
+        assert check_intersection == queryable_fields
```
renaudjester, line 203: Maybe add a test that does snapshots of the available fields, I don't know how easy it would be to get them 

But when you have them you can do: asset sorted(list(available_fields))

uriii3: Does it make any sense what I'm doing with the `get_queryable` fields option anyway, then? I actually saw that it wasn't doing much, so maybe the snapshot is the best and it works fine.

uriii3:
Note to self: snapshots don't work with python versions that are too old, keep them in the range they say they work on (for now, up to 3.10). link to [pytest-snapshot](https://pypi.org/project/pytest-snapshot/).

renaudjester:
> Note to self: snapshots don't work with python versions that are too old, keep them in the range they say they work on (for now, up to 3.10). link to [pytest-snapshot](https://pypi.org/project/pytest-snapshot/).

We are not using python-snapshot, we are using [syrupy](https://github.com/syrupy-project/syrupy)

renaudjester (approved):
Super good idea the snapshots, way better

PR #232, opened 2024-11-26
chore: add raise for status for better error messages

renaudjester:
fix [CMT-149](https://cms-change.atlassian.net/browse/CMT-149)

----
📚 Documentation preview 📚: https://copernicusmarine--232.org.readthedocs.build/en/232/

### DIFF
--- a/copernicusmarine/catalogue_parser/catalogue_parser.py
+++ b/copernicusmarine/catalogue_parser/catalogue_parser.py
@@ -17,12 +17,9 @@ from copernicusmarine.catalogue_parser.models import (
     get_version_and_part_from_full_dataset_id,
 )
 from copernicusmarine.core_functions.sessions import (
-    get_configured_requests_session,
+    JsonParserConnection as CatalogParserConnection,
 )
-from copernicusmarine.core_functions.utils import (
-    construct_query_params_for_marine_data_store_monitoring,
-    run_concurrently,
-)
+from copernicusmarine.core_functions.utils import run_concurrently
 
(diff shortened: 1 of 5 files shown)


copernicusmarine/core_functions/credentials_utils.py
```diff
@@ -342,9 +342,11 @@ def _check_credentials_with_old_cas(username: str, password: str) -> bool:
         f"https://cmems-cas.cls.fr/cas/login?service={service}"
     )
     conn_session = get_configured_requests_session()
+    logger.debug(f"GETing {cmems_cas_login_url}...")
```
uriii3, line 345: weird naming, or is it in purpose?

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -17,12 +17,9 @@
     get_version_and_part_from_full_dataset_id,
 )
 from copernicusmarine.core_functions.sessions import (
-    get_configured_requests_session,
-)
-from copernicusmarine.core_functions.utils import (
-    construct_query_params_for_marine_data_store_monitoring,
-    run_concurrently,
+    JsonParserConnection as CatalogParserConnection,
```
uriii3, line 20: seems weird the renaming, why not just change the name of the variable in the code?

copernicusmarine/core_functions/credentials_utils.py
renaudjester: It's because they did it like this for the POST request, it's to say that we are doing a GET http request

PR #240, opened 2024-12-02
chore: syncing parts allowed

uriii3:
When we have more than one part, allow still the syncing of the options. Correcting [CMT-151](https://cms-change.atlassian.net/browse/CMT-151) also for version 2 of the toolbox (and even allowing it for several parts).

Currently, the tests are done on the staging area, as in PRD the datasets have been removed for the moment. We will need to update this.

----
📚 Documentation preview 📚: https://copernicusmarine--240.org.readthedocs.build/en/240/

### DIFF
--- a/copernicusmarine/core_functions/get.py
+++ b/copernicusmarine/core_functions/get.py
@@ -152,7 +152,6 @@ def _run_get_request(
         get_request.force_dataset_part,
         None,
         CommandType.GET,
-        get_request.index_parts,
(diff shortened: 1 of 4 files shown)


renaudjester (approved):
Please just verify if in the docstrings and the documentation we don't have a line about this (like in the sync paragraphe: doesn't work if dataset have more than one part)

uriii3:
Indeed we have such line. I will delete it now.

PR #243, opened 2024-12-04
chore: solving chunk search for omi

uriii3:
Fixing bug [CMT-161](https://cms-change.atlassian.net/browse/CMT-161), where because the dataset is OMI, it has a values set, but those are not numbers but rather regions -> hard to estimate the size then. I just skip them, as if the values were not correctly set.

----
📚 Documentation preview 📚: https://copernicusmarine--243.org.readthedocs.build/en/243/


uriii3:
It looks weird because my pre-commit didn't trigger, I will investigate it now.

PR #245, opened 2024-12-04
chore: documentation domains & dependencies

uriii3:
Fixing tickets [CMT-158](https://cms-change.atlassian.net/jira/software/c/projects/CMT/boards/24/backlog?selectedIssue=CMT-158) and also [CMT-156](https://cms-change.atlassian.net/browse/CMT-156). 

Mainly it is adding the domains that the toolbox uses (not anymore found by my in the code) and also the dependencies (see that we have no 'optional' dependencies, as it was something that is on our roadmap to be checked, how to separate the dependencies).

----
📚 Documentation preview 📚: https://copernicusmarine--245.org.readthedocs.build/en/245/


doc/installation.rst
```diff
@@ -159,3 +178,13 @@ If you want to use it please make sure to install the ``netcdf4`` library:
 .. note::
 
     The image of the toolbox should already have the ``netcdf4`` library installed.
+
+
+Domains required by the Copernicus Marine toolbox
+********************************************************
+To be able to debug, we provide the domains that the toolbox uses when running:
```
renaudjester, line 185:
```suggestion
To be able to use the Copernicus Marine services, you need to be able to access those domains: 
```

pyproject.toml
```diff
@@ -26,6 +26,7 @@ numpy = ">=1.23.0"
 pendulum = "^3.0.0"
 pydantic = "^2.9.1"
 h5netcdf = "^1.4.0"
+types-toml = "^0.10.8.20240310"
```
uriii3, line 29: how do I know that I strictly need this here? I'm not sure anymore

renaudjester: No you don't need them here! They shouldn't be here because they are only used for testing

renaudjester: so it should be in the dev dependencies (`poetry add --dev types-toml` or equivalent)

tests/__snapshots__/test_dependencies_updates.ambr
```diff
@@ -0,0 +1,29 @@
...
+    'xarray': '>=2023.4.0',
+    'zarr': '>=2.13.3',
+  })
+# ---
+# name: TestDependeciesUpdates.test_update_dependencies.1
+  dict({
+    'ipython': '^8.14.0',
+    'pre-commit': '^2.20.0',
+    'types-requests': '^2.27.11',
+  })
```
renaudjester, line 28: I wouldn't track that since it's dev dependencies

pyproject.toml
uriii3: Okay: looks like the poetry way of handling dependencies is being updated, so I will put this new one in the updated form: [link here](https://python-poetry.org/docs/managing-dependencies/).

conda_environment_test.yaml
```diff
@@ -10,3 +10,5 @@ dependencies:
   - pip:
     - pytest-order==1.2.1
     - freezegun==1.5.1
+    - toml==0.10.2
+    - types-toml==0.10.8.20240310
```
renaudjester, line 14: Not sure we need it :D

pyproject.toml
```diff
@@ -35,6 +35,9 @@ ipython="^8.14.0"
 [tool.poetry.scripts]
 copernicusmarine = 'copernicusmarine.command_line_interface.copernicus_marine:command_line_interface'
 
+[tool.poetry.group.dev.dependencies]
+types-toml = "^0.10.8.20240310"
```
renaudjester, line 39: Ok so the problem is that now you have created a **group** for this, whereas there is a dev dependencies that exists already (`[tool.poetry.dev-dependencies]`) it could be coming from the difference of versions from the poetry you are using and the poetry used before to create the file

Anyway it would be nice to homogenize this: all dev deps in `[tool.poetry.dev-dependencies]` or all dev deps in `[tool.poetry.dev-dependencies]`

conda_environment_test.yaml
uriii3: we actually do hehe (check tests from commit [6c9d218](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/245/commits/6c9d21876564fe85833de5f89bffcc540739b5d1))

renaudjester (approved):

PR #248, opened 2024-12-09
chore: updating attrs

uriii3:
We discovered that for some datasets the attributes were not set in the Lon and lat dimensions. We are solving it with this PR.

----
📚 Documentation preview 📚: https://copernicusmarine--248.org.readthedocs.build/en/248/


copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -511,6 +511,25 @@ def _update_dataset_coordinate_attributes(
                     coordinate_attributes.remove("units")
                 elif coordinate_label in ["depth", "elevation"]:
                     coordinate_attributes.append("positive")
+                if (
+                    coord.attrs == {}
+                ):  # specific dataset case without lon and lat attrs
```
renaudjester, line 516: If you want to add a comment remember to comment the "why" and not the "what" because reading the code we see that it is a case where the lat lon are don't have attributs

I would put something like: 
```suggestion
                    coordinate_attributes.append("positive")
                # TODO: delete this when fixed on ARCO processor side
                # example dataset: klshdflksjdf
                if (
                    coord.attrs == {}
                ): 
```

copernicusmarine/download_functions/subset_xarray.py
```diff
@@ -511,6 +511,25 @@ def _update_dataset_coordinate_attributes(
                     coordinate_attributes.remove("units")
                 elif coordinate_label in ["depth", "elevation"]:
                     coordinate_attributes.append("positive")
+                if (
+                    coord.attrs == {}
```
renaudjester, line 515:
```suggestion
                    attrs == {}
```

copernicusmarine/download_functions/subset_xarray.py
renaudjester, line 1: Maybe you can add a test?

renaudjester (approved):

PR #264, opened 2025-01-09
chore: trigger the binaries after the release

renaudjester:
We noticed the the creation of the binaries might finish before the release is done so trying to do the binaries after the release

----
📚 Documentation preview 📚: https://copernicusmarine--264.org.readthedocs.build/en/264/

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -3,7 +3,6 @@ name: Create Binaries MacOS and Linux
 on:
   workflow_call:
 
-
 jobs:
   build:
     strategy:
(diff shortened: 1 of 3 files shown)


uriii3 (approved):
Looks amazing and supercool! We don't have anyway to test it until release though, no? Should we also add a manual way to do it, just in case? Or with something like 're-run all jobs' will be enough?

renaudjester:
> Looks amazing and supercool! We don't have anyway to test it until release though, no? Should we also add a manual way to do it, just in case? Or with something like 're-run all jobs' will be enough?

Agreed that the testing has been limited! But we should be able to rerun the tests individually (job per job)

PR #274, opened 2025-01-20
manually add dependency deprecated when building binaries

uriii3:
Trying to solve the latest crash with the binaries.

----
📚 Documentation preview 📚: https://copernicusmarine--274.org.readthedocs.build/en/274/


uriii3:
https://github.com/pyinstaller/pyinstaller/issues/8993

renaudjester (approved):

PR #275, opened 2025-01-23
better estimating the file size

uriii3:
We discovered some incongruence between the size of the dataset and our estimation. Although it seems that the proposed new method is not bullet-proof either, it looks like it will be more consistent in the estimation.

Trying to solve issues [CMT-175](https://cms-change.atlassian.net/browse/CMT-175) and [CMT-171](https://cms-change.atlassian.net/browse/CMT-171).

----
📚 Documentation preview 📚: https://copernicusmarine--275.org.readthedocs.build/en/275/

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -50,14 +50,15 @@ def main_checks_when_file_is_downloaded(
     response: dict,
 ):
     size_variance = 0.2
+    offset_size = 0.05  # small datasets are hard to predict
     file_size = os.path.getsize(file_path)
     assert (
         file_size / 1048e3
-        <= response["file_size"] * (1 + size_variance) + 0.05
+        <= response["file_size"] * (1 + size_variance) + offset_size
     )
     assert (
         file_size / 1048e3
-        >= response["file_size"] * (1 - size_variance) - 0.05
+        >= response["file_size"] * (1 - size_variance) - offset_size
     )
     assert response["file_size"] <= response["data_transfer_size"]
     return
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -49,7 +49,7 @@ def main_checks_when_file_is_downloaded(
     file_path: pathlib.Path,
     response: dict,
 ):
-    size_variance = 0.4
+    size_variance = 0.5
     file_size = os.path.getsize(file_path)
     assert file_size / 1048e3 <= response["file_size"] * (1 + size_variance)
     assert file_size / 1048e3 >= response["file_size"] * (1 - size_variance)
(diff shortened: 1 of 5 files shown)


uriii3:
Some examples with it's discrepancies:

**Example 1:**
```
copernicusmarine subset --dataset-id cmems_mod_glo_phy_myint_0.083deg_P1D-m --variable thetao --start-datetime 2024-10-22T00:00:00 --end-datetime 2024-10-22T00:00:00 --minimum-longitude -180 --maximum-longitude 179.9166717529297 --minimum-latitude -80 --maximum-latitude 90 --minimum-depth 0.49402499198913574 --maximum-depth 5727.9169921875
```
Estimated: 3365 MB
Real final size: 881 MB -> this is the weirdest case, maybe a thorough investigation should be done.

**Example 2:**
```
copernicusmarine subset --dataset-id cmems_mod_glo_phy_anfc_0.083deg-climatology-uncertainty_P1M-m -t 1990
```
Estimated: 1211 MB -> this one has improved from an old estimation of 2400MB!! (so we are better now)
Real final size: 1270 MB 

**Example 3:**
```
copernicusmarine subset -i baltic_omi_health_codt_volume -t 2020     
```
Estimated: 16 kB -> this has improved from 5kb, which means we are better now!
Real final size: 23 kB

uriii3:
Maybe (seeing that all checks have passed) we could add some checks into files that we might already be downloading, to check that the estimated size and the final size doesn't diverge a threshold value? what do you think @renaudjester ?

renaudjester (approved):
So it's improved but not perfect right? And it doesn't solve all the use cases of CMT-175 I guess

Theophile-Varnier (approved):

tests/test_utils.py
```diff
@@ -41,3 +43,22 @@ def execute_in_terminal(
...
+    assert (
+        file_size / 1048e3
+        <= response["file_size"] * (1 + size_variance) + offset_size
+    )
+    assert (
+        file_size / 1048e3
+        >= response["file_size"] * (1 - size_variance) - offset_size
+    )
+    assert response["file_size"] <= response["data_transfer_size"]
+    return
```
Theophile-Varnier, line 64: do you need this return?

uriii3: I don't know actually

uriii3: I checked it and it is more of a [style thing](https://realpython.com/python-return-statement/#using-the-python-return-statement-best-practices). I also checked the repo and it looks like the 3 versions are on it. I'm used to this but we can standardise the use if you want to!

Theophile-Varnier: vale then good for me!

renaudjester (approved):

PR #278, opened 2025-02-06
dynamically use version in binaries links

Theophile-Varnier:
Fixes [CMT-197](https://cms-change.atlassian.net/browse/CMT-197)

----
📚 Documentation preview 📚: https://copernicusmarine--278.org.readthedocs.build/en/278/

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -19,3 +19,4 @@ ignore =
 per-file-ignores =
     # __init__ files
     dataset_product_mapping.py: E501
+    conf.py: E501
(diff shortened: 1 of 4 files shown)


uriii3 (commented):
hopefully if we add `toml` to the dependencies in the documentation we can see it working and then it will be way easier to review, although looks good!

uriii3 (changes requested):
There appears the #noqa E501 in the documentation hehehe
image

renaudjester (approved):

uriii3 (approved):

PR #279, opened 2025-02-11
chore: add changelog

renaudjester:
Change log for version 2.0.1

----
📚 Documentation preview 📚: https://copernicusmarine--279.org.readthedocs.build/en/279/

### DIFF
--- a/doc/changelog/v201.rst
+++ b/doc/changelog/v201.rst
@@ -1,0 +1,24 @@
+Release 2.0.1
+=====================
+
+Patch release with bug fixes and improvements.
+
+General
+''''''''
+
+Fixes
+------------------
+
+* Fixed some typos in the documentation have been fixed.
+* Fixed broken links to binaries in the documentation when releasing a new version. They now point to the same version as the version of the documentation.
+
+Subset
+''''''''
+
+Fixes
+------------------
+
+* Fixed a bug where input just outside the range of the dataset would not be handled correctly.
+* Fixed a bug when calculating the 'nearest' point for a dataset, it would not always return the correct point.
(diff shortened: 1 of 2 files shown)


doc/changelog/v201.rst
```diff
@@ -0,0 +1,24 @@
...
+* Fixed broken links to binaries in the documentation when releasing a new version. They now point to the same version as the version of the documentation.
+
+Subset
+''''''''
+
+Fixes
+------------------
+
+* Fixed a bug where input just outside the range of the dataset would not be handled correctly.
+* Fixed a bug when calculating the 'nearest' point for a dataset, it would not always return the correct point.
```
uriii3, line 22:
```suggestion
* Fixed a bug when using the 'nearest' method for a subset selection, it would not always return the correct extreme points.
```

doc/changelog/v201.rst
```diff
@@ -0,0 +1,24 @@
...
+Patch release with bug fixes and improvements.
+
+General
+''''''''
+
+Fixes
+------------------
+
+* Fixed some typos in the documentation.
+* Fixed broken links to binaries in the documentation when releasing a new version. They now point to the same version as the version of the documentation.
```
uriii3, line 13: I don't fully understand the second part of the phrase, it is properly written? What do you mean?

uriii3 (approved):

doc/changelog/v201.rst
renaudjester: So the documentation has a version so the binaries will point to this specific version, before they were pointing to "latest" if I am not wrong

uriii3: Correctly, perfect!

PR #283, opened 2025-02-18
release zarr version constraint

Theophile-Varnier:
Fixes [CMT-188](https://cms-change.atlassian.net/browse/CMT-188)

- Implementation of a new custom Zarr store according to the Python Zarr v3 library interface
- Selection of the custom Zarr store depending on the current Python Zarr version
- There are no automated tests for now, as they will be added during the realisation of [CMT-186](https://cms-change.atlassian.net/browse/CMT-186)


Theophile-Varnier:
Right now it seems we are facing this issue: https://github.com/pydata/xarray/issues/10032

rbanick:
Setting compression with `compressor = (zarr.codecs.BloscCodec(cname="lz4")` (or whatever `cname` you would like to use) effectively fixes this issue.

Does Copernicus Marine have a timeline for implementing this and other Zarr3 fixes so the API's v3 capabilities function? I'm trying to avoid bifurcating my stack to handle Copernicus Marine with Zarr v2 and all other datasets with Zarr v3...

copernicusmarine/download_functions/download_zarr.py
```diff
@@ -10,9 +10,13 @@
 
 if zarr.__version__.startswith("2"):
     from zarr.storage import DirectoryStore
+
+    ZARR_FORMAT = None
 else:
     from zarr.storage import LocalStore as DirectoryStore
 
+    ZARR_FORMAT = 2
+
```
uriii3, line 19: isn't it a bit weird that if it starts with 2 it is set to None and otherwise, set to 2? I would prefer a simple variable, "ZARR _UPDATED" which is just True or false, and in the case it is true, just introduce 2 in the call below? Or that is complicating code and it is easier to see that the 2 in the function below will come from this?

uriii3 (approved):

copernicusmarine/download_functions/download_zarr.py
renaudjester: Not sure it would make it clearer 🤔 

What if I add comments here about why we are doing this? would it be enough?

PR #287, opened 2025-03-04
feat: add platformids subsetting

renaudjester:
Add platform ids subsetting and add some information in the describe to help user identify platform ids name


tests/test_sparse_subset.py
```diff
@@ -0,0 +1,63 @@
+from tests.test_utils import execute_in_terminal
+
+# TODO: maybe reduce the size of the request
+BASIC_COMMAND = [
+    "copernicusmarine",
+    "subset",
+    "--dataset-id",
+    "cmems_obs-ins_arc_phybgcwav_mynrt_na_irr",
+    "--dataset-part",
+    "history",
```
uriii3, line 10: maybe a stupid question, but are INSITUs always a `history dataset_part` ? How do we communicate that to users?

uriii3 (commented):
as a general thing, it looks like both this branch and the stereographic branch rely on knowing the 'dataset_part' that the user wants to retrieve (either the insets or the stereographic or it is not forced). Maybe we could find a way to standardise this? I feel like right know we are both passing some variables to make it work, but not sure if they are very unified.

PR #290, opened 2025-03-13
Updating ubuntu runner

uriii3:
From https://github.com/actions/runner-images/issues/11101 we know we will need to update the runner.

I finally chose to just update the runner rather than just deleting it, so we keep having 2 versions for ubuntu.

----
📚 Documentation preview 📚: https://copernicusmarine--290.org.readthedocs.build/en/290/


uriii3:
(I ran the binaries test manually and they seem to work!)

Theophile-Varnier (approved):

PR #294, opened 2025-03-20
feat: accepts real platform names without the types

renaudjester:
Improvement to the current sparse subsetting

Current platform ids on the metadata are with the type of the platform: `<platform_name>___<type>`

ex: `B-Sulafjorden___MO` 

Now the toolbox accepts the input `B-Sulafjorden` and retrieve all the platforms with this name 

Also sets `csv` as the default format.

----
📚 Documentation preview 📚: https://copernicusmarine--294.org.readthedocs.build/en/294/


doc/conf.py
```diff
@@ -78,6 +78,10 @@
 
 nb_execution_mode = "off"
 
+# -- Custom RST epilog --------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog
+# Allows us to have the version dynamically updated in the documentation
+
 rst_epilog = """
```
uriii3, line 85: ? what is that for?

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -197,18 +209,34 @@ def _get_user_configuration(username: str) -> UserConfiguration:
...
+) -> list[str]:
+    platforms_to_subset = []
     if platform_ids:
         platforms_names = get_platforms_names(metadata_url, user_configuration)
         if not platforms_names:
             raise NotEnoughPlatformMetadata()
+        platforms_names_with_types: set[str] = set()
+        platforms_without_types_mapping: dict[str, list] = defaultdict(list)
+        for platform_name in platforms_names:
+            platform_name_without_type = platform_name.split("___")[0]
```
uriii3, line 226: is this the only ay that platforms will be spliced with their types? is this 'standard', somehow?

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -197,18 +209,34 @@ def _get_user_configuration(username: str) -> UserConfiguration:
...
+            platforms_without_types_mapping[platform_name_without_type].append(
+                platform_name
+            )
+            platforms_names_with_types.add(platform_name)
```
uriii3, line 230: why is this different than `platforms_names` ?  it is not in place so you are duplicating a list, no?

uriii3 (approved):

copernicusmarine/download_functions/download_sparse.py
renaudjester: Yes it's created when doing the metadata so it should be

copernicusmarine/download_functions/download_sparse.py
renaudjester: yes kind of but I am transforming into a set so that the `in` after should be way faster

doc/conf.py
renaudjester: Just some info :D

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -45,15 +47,16 @@ def download_sparse(
...
-        metadata_url,
-        service,
-        user_configuration,
-    )
-    extension_file = get_file_extension(
-        subset_request.file_format or "parquet"
-    )
+    if subset_request.platform_ids:
+        platform_ids = _get_plaform_ids_to_subset(
+            subset_request.platform_ids or [],
```
Theophile-Varnier, line 52: is this `or []` still necessary?

Theophile-Varnier (approved):

PR #295, opened 2025-03-20
fix: only ascii characters in documentation

renaudjester:
fix [CMT-206](https://cms-change.atlassian.net/browse/CMT-206)

Based on the list of ASCII characters, delete the special characters in the documentation of the functions: https://www.ascii-code.com/

----
📚 Documentation preview 📚: https://copernicusmarine--295.org.readthedocs.build/en/295/


tests/__snapshots__/test_help_command_interface.ambr
```diff
@@ -340,7 +340,7 @@
     '                                  be inside the requested range. If ``strict-',
     '                                  inside``, the selection retrieved will be',
     '                                  inside the requested range, and an error',
-    "                                  will be raised if the values don't exist. If",
+    "                                  will be raised if the values don\\'t exist. If",
```
uriii3, line 343: this looks weird!

tests/__snapshots__/test_help_command_interface.ambr
```diff
@@ -392,13 +392,13 @@
...
     '                                  as a string. Caution, user has to replace',
-    "                                  double quotes \\xe2\\x80\\x9c with single quotes ' in the",
+    '                                  double quotes " with single quotes \\\' in the',
     '                                  request.',
     '  --dry-run                       If True, runs query without downloading',
     '                                  data.',
     '  -r, --response-fields TEXT      List of fields to include in the query',
     '                                  metadata. The fields are separated by a',
-    "                                  comma.To return all fields, use 'all'.",
+    "                                  comma.To return all fields, use \\'all\\'.",
```
uriii3, line 401:
```suggestion
    "                                  comma. To return all fields, use \\'all\\'.",
```
Can you add this change, also? I just saw it (definitely not here but in the right place in the code)

tests/__snapshots__/test_help_command_interface.ambr
uriii3, line 1: why did this change? What happened that made this update like this? Did you run snapshot-update?

uriii3:
We implemented this so the installation of the binaries didn't provoke any problem to the users: https://github.com/orgs/pyinstaller/discussions/9065

tests/__snapshots__/test_help_command_interface.ambr
renaudjester: I ran snapshot update indeed! I think it's because to save the result the library is escaping some characters and since we change the characters it needs to escape differently I suppose

uriii3: But this will not be the output in the terminal, no?

uriii3 (approved):

tests/__snapshots__/test_help_command_interface.ambr
renaudjester: No I don't think so

PR #296, opened 2025-03-20
feat: proper names for sparse subset

renaudjester:
Now the name should be a similar format as the dataset for zarr formatted data

The difference is that the name reflects the requested value instead of the real value that user will get

TODO:

- [ ] Add documentation for this difference

### DIFF
--- a/copernicusmarine/download_functions/download_sparse.py
+++ b/copernicusmarine/download_functions/download_sparse.py
@@ -32,7 +32,10 @@ from copernicusmarine.core_functions.utils import (
     construct_query_params_for_marine_data_store_monitoring,
     get_unique_filepath,
 )
-from copernicusmarine.download_functions.utils import get_file_extension
+from copernicusmarine.download_functions.utils import (
+    build_filename_from_request,
+    get_file_extension,
+)
 
 logger = logging.getLogger("copernicusmarine")
 
@@ -44,6 +47,7 @@ def download_sparse(
     subset_request: SubsetRequest,
     metadata_url: str,
     service: CopernicusMarineService,
+    axis_coordinate_id_mapping: dict[str, str],
     disable_progress_bar: bool,
 ) -> ResponseSubset:
     user_configuration = _get_user_configuration(username)
@@ -56,10 +60,19 @@ def download_sparse(
         )
     else:
         platform_ids = []
+    variables = subset_request.variables or [
+        variable.short_name for variable in service.variables
+    ]
+
     extension_file = get_file_extension(subset_request.file_format)
     filename = pathlib.Path(
         subset_request.output_filename
-        or f"{subset_request.dataset_id}_subset{extension_file}"
+        or build_filename_from_request(
+            subset_request,
+            variables,
+            platform_ids,
+            axis_coordinate_id_mapping,
+        )
     )
 
     if filename.suffix != extension_file:
@@ -70,9 +83,7 @@ def download_sparse(
     )
     if not subset_request.overwrite and not subset_request.skip_existing:
         output_path = get_unique_filepath(output_path)
-    variables = subset_request.variables or [
-        variable.short_name for variable in service.variables
-    ]
+
     response = ResponseSubset(
         file_path=output_path,
         output_directory=subset_request.output_directory,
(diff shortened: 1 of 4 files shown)


copernicusmarine/download_functions/download_sparse.py
```diff
@@ -44,6 +47,7 @@ def download_sparse(
     subset_request: SubsetRequest,
     metadata_url: str,
     service: CopernicusMarineService,
+    axis_coordinate_id_mapping: dict[str, str],
```
uriii3, line 50: I'm actually a bit happy that this is useful also for this hehe 🤗

copernicusmarine/download_functions/utils.py
```diff
@@ -143,6 +140,79 @@ def _build_filename_from_dataset(
...
+    platform_ids: list[str],
+    axis_coordinate_id_mapping: dict[str, str],
+) -> str:
+    """
+    In the sparse dataset case we don't have the dataset to build the filename from.
+    """  # noqa
+
+    dataset_variables = "-".join(variables)
+    dataset_variables = (
+        "multi-vars" if len(variables) > 15 else dataset_variables
```
uriii3, line 155: 15 variables is a lot of variables... wouldn't it be best to set a lower threshold?

copernicusmarine/download_functions/utils.py
```diff
@@ -143,6 +140,79 @@ def _build_filename_from_dataset(
...
+    platform_ids_text = (
+        "multi-platforms" if len(platform_ids) > 15 else platform_ids_text
+    )
+    longitudes = None
+    if "x" in axis_coordinate_id_mapping:
+        if axis_coordinate_id_mapping["x"] == "longitude":
+            longitudes = _format_longitudes(
+                request.minimum_x, request.maximum_x
+            )
+        if axis_coordinate_id_mapping["x"] == "x":
```
uriii3, line 167: do we have insitu with original data? 😮

copernicusmarine/download_functions/utils.py
```diff
@@ -143,6 +140,79 @@ def _build_filename_from_dataset(
...
+                dataset_variables,
+                platform_ids_text,
+                longitudes,
+                latitudes,
+                depths,
+                datetimes,
+            ],
+        )
+    )
+    filename = filename if len(filename) < 250 else filename[250:]
```
uriii3, line 211: don't we maybe prefer to cut the variables names than just the requested info? whatever you prefer!

uriii3 (approved):

copernicusmarine/download_functions/utils.py
```diff
@@ -143,6 +140,79 @@ def _build_filename_from_dataset(
...
+    dataset_variables = "-".join(variables)
+    dataset_variables = (
+        "multi-vars" if len(variables) > 15 else dataset_variables
+    )
+    platform_ids_text = "-".join(platform_ids)
+    platform_ids_text = (
+        "multi-platforms" if len(platform_ids) > 15 else platform_ids_text
+    )
+    longitudes = None
+    if "x" in axis_coordinate_id_mapping:
```
Theophile-Varnier, line 162: Isn't this part a duplicate of what we have above?

copernicusmarine/download_functions/utils.py
renaudjester: It's been always like this 🤔

copernicusmarine/download_functions/utils.py
renaudjester: Nope but I thought that might as well have it since the code is already written

copernicusmarine/download_functions/utils.py
renaudjester: if you mean that this function is very similar to the other one then yes! 

The difference is that the one below is using the xarray.Dataset (lazy loading) to get the exact data we are gonna get from the subset! here we are using the request so it doesn't reflect what is inside the result

copernicusmarine/download_functions/utils.py
renaudjester: Same it's always been like this!

Theophile-Varnier (approved):

PR #299, opened 2025-03-26
fix: update docs

renaudjester:
From the comments about the alpha version

----
📚 Documentation preview 📚: https://copernicusmarine--299.org.readthedocs.build/en/299/

### DIFF
--- a/doc/usage/subset_using_polygons.ipynb
+++ b/doc/usage/subset_using_polygons.ipynb
@@ -5,7 +5,8 @@
    "id": "e2ddf5cd",
    "metadata": {},
    "source": [
-    "# Polygon-based data selection"
+    "(polygons-based-data-selection)=\n",
+    "# Polygons-based data selection"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


doc/usage/subset_using_polygons.ipynb
```diff
@@ -5,7 +5,8 @@
    "id": "e2ddf5cd",
    "metadata": {},
    "source": [
-    "# Polygon-based data selection"
+    "(polygons-based-data-selection)=\n",
+    "# Polygons-based data selection"
    ]
```
uriii3, line 10: do you change this manually? and where are they pointed, though? I thought we wanted them to be able to refer to them from the changelog, no?

uriii3 (approved):

doc/usage/subset_using_polygons.ipynb
renaudjester: Yep but the change log is only in the pre-release branch so I will add it later! 

And you can use is the same as the reference we do in rst (the `.. _some-ref:`)

PR #301, opened 2025-03-28
[fix] improve naming of stereo datasets

uriii3:
We recently added datasets with 'yc' and 'xc' components, and after checking the naming I corrected them. Now, for example:

`copernicusmarine subset --dry-run --dataset-part originalGrid -t 2021 -i cmems_obs-si_arc_physic_nrt_1km-grl_P1D-irr`

Outputs this filename:

`cmems_obs-si_arc_physic_nrt_1km-grl_P1D-irr_multi-vars_-923125.69x_1307874.25X_-3846677.75y_-529677.75Y_2021-01-01-2025-03-27.nc`

----
📚 Documentation preview 📚: https://copernicusmarine--301.org.readthedocs.build/en/301/

### DIFF
--- a/copernicusmarine/download_functions/download_zarr.py
+++ b/copernicusmarine/download_functions/download_zarr.py
@@ -137,6 +137,7 @@ def download_dataset(
         dataset_id,
         file_format,
         axis_coordinate_id_mapping,
+        geographical_parameters,
     )
     output_path = pathlib.Path(output_directory, filename)
     final_result_size_estimation = get_approximation_size_final_result(
--- a/copernicusmarine/download_functions/utils.py
+++ b/copernicusmarine/download_functions/utils.py
@@ -47,6 +47,7 @@ def get_filename(
     dataset_id: str,
     file_format: FileFormat,
     axis_coordinate_id_mapping: dict[str, str],
+    geographical_parameters: GeographicalParameters,
 ) -> str:
     if filename:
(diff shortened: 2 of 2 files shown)


uriii3:
I got to say that I don't know if I like the 'depths' part... it looks like it will break too for some datasets. Should I take into account the different 'depth', 'elevation' and 'depthc' ?  any other name for that? (but I don't really like to focus on the name to set the filename...)

uriii3:
In branch v2.0.1:
Command: `copernicusmarine subset --dataset-id cmems_mod_ibi_phy_my_0.083deg-3D_P1D-m --variable thetao --variable so --start-datetime 2021-01-01 --end-datetime 2021-01-03 --minimum-longitude 0.0 --maximum-longitude 0.1 --minimum-latitude 28.0 --maximum-latitude 28.1 --dry-run --vertical-axis elevation`

Retrieves: `cmems_mod_ibi_phy_my_0.083deg-3D_P1D-m_thetao-so_0.08E_28.00N-28.08N_5698.06-0.51m_2021-01-01-2021-01-03.nc`

uriii3:
It introduces (this branch) a breaking change, as it is proposing to update the naming convention we have for the depths. If elevation, it proposes to invert the order, to be able to put the abs min to the abs max... We will first merge the branch that only corrects the bugs (https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/303) and then maybe in the future we also consider merging this update.

PR #303, opened 2025-03-31
[fix] Update way to handle naming

uriii3:
Updating a way to handle naming. Correcting also a mini small bug but that it seemed quite relevant 🥺 .

Let me know if you think the correction of the bug is also breaking change... I consider it to be relevant enough but it might be maybe!

----
📚 Documentation preview 📚: https://copernicusmarine--303.org.readthedocs.build/en/303/


uriii3:
Not a breaking change, as it only changes the status of alpha version.

renaudjester (approved):

Theophile-Varnier (approved):

PR #307, opened 2025-04-02
[fix] Improving stereo docs

uriii3:
Revising the documentation once again to create an even more comprehensive an accurate file for users to approach it.


doc/usage/stereo_subset.ipynb
```diff
@@ -4,53 +4,56 @@
...
     "\n",
-    "Maintaining data in the original-grid format, rather than only providing longitude and latitude coordinates, is valuable because it preserves the original structure and integrity of the dataset. This approach eliminates the need for reprojection or interpolation steps required when transforming data into geographic coordinates. Additionally, it provides convenience for users who prefer working directly with the original-grid coordinates.\n",
+    "Maintaining data in the original-grid format, rather than only providing longitude and latitude coordinates, offers significant benefits:\n",
+    "- It preserves the dataset's original structure and integrity\n",
+    "- It eliminates the requirement for reprojection or interpolation when converting to geographic coordinates \n",
```
gkoenig-mercator, line 19: I am not sure about this point. I mean, we had the data in geographical coordinates in the first place, so that requirement never really existed, did it?

doc/usage/stereo_subset.ipynb
```diff
@@ -4,53 +4,56 @@
...
+    "(stereographic-subsetting-page)=\n",
+    "# Subsetting ARCO Data with Original-grid"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The (longitude, latitude) projection is widely used. However, when going to places near the poles, the Arctic or the Antartic regions, there are other sets of coordinates which are used and that can be useful. One of such projections is the stereographic projection, although there are others. The exact projection used in each dataset is explained in the product, you can check the [Products in Copernicus Marine Data Store](https://data.marine.copernicus.eu/products).\n",
+    "The (longitude, latitude) projection is widely used. However, for regions near the poles - the Arctic or the Antarctic - other coordinate systems are employed, including the stereographic projection among others. The specific projection used for each dataset is detailed in the respective product documentation, available through the [Copernicus Marine Data Store](https://data.marine.copernicus.eu/products).\n",
```
gkoenig-mercator, line 15: "The geographical grid or coordinate system with latitude and longitude is widely used" I would say

doc/usage/stereo_subset.ipynb
```diff
@@ -59,29 +62,32 @@
    "source": [
     "For a given dataset and version, the data in stereographic projection is in a different part.\n",
     "\n",
-    "This is why, when subsetting data based on the original-grid, it is necessary to explicitly specify the dataset part. In the command-line interface, this is achieved by using the option `--dataset-part originalGrid`. Similarly, in the Python interface, you would specify this parameter as follows: `dataset_part=\"originalGrid\"`."
+    "Consequently, when subsetting original-grid data, you must explicitly specify the dataset part:\n",
+    "\n",
+    "- Command-line interface: Use ``--dataset-part originalGrid``\n",
+    "- Python interface: Specify ``dataset_part=\"originalGrid\"``"
```
gkoenig-mercator, line 68: It would be nice to add a line saying "If no parameter is specified, the datasets in geographical coordinates will be downloaded."

doc/usage/stereo_subset.ipynb
```diff
@@ -59,29 +62,32 @@
...
-    "## Python interface and plot comparison"
+    "## Python Interface and Plot Comparison"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To better understand how the data is retrieved, let's see an example where similar regions are plotted with the toolbox. We will first present the way that it would work with longitude and latitude, and then present the new option with the original-grid projection. "
+    "To demonstrate the data retrieval process, we provide an example comparing plots of similar regions using the toolbox. We first illustrate the longitude/latitude approach, followed by the original-grid projection method."
```
gkoenig-mercator, line 82: "We first illustrate how to download datasets in geographical coordinates" would sound a little bit nicer

doc/usage/stereo_subset.ipynb
```diff
@@ -901,12 +907,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### With stereographic projection\n",
+    "### With Stereographic Projection\n",
     "\n",
-    "- Instead of setting limits using ``minimum_longitude`` and ``minimum_latitude`` (and their maximum counterparts), we now define spatial boundaries directly in projected coordinates using ``minimum_x``, ``maximum_x``, ``minimum_y`` and ``maximum_y``.\n",
-    "- we need to specify the dataset part, as the \"originalGrid\" part as explained at the start of the notebook.\n",
+    "- Spatial boundaries are defined using projected coordinates (``minimum_x``, ``maximum_x``, ``minimum_y``, ``maximum_y``) rather than geographic coordinates (``minimum_longitude``, ``minimum_latitude``, etc.)\n",
```
gkoenig-mercator, line 912: "Spatial boundaries are defined onto the stereographic grid (``minimum_x``, ``maximum_x``, ``minimum_y``, ``maximum_y``) rather than the geographical one..."

doc/usage/stereo_subset.ipynb
```diff
@@ -1897,169 +1905,37 @@
...
-     "text": [
-      "ERROR - 2025-03-13T08:39:33Z - Lon lat subset not available in original grid datasets: You cannot specify longitude and latitude when using the 'originalGrid' dataset part yet. Try using ``--minimum-x``, ``--maximum-x``, ``--minimum-y`` and ``--maximum-y``.\n"
-     ]
-    }
-   ],
-   "source": [
-    "!copernicusmarine subset -i cmems_mod_arc_bgc_my_ecosmo_P1D-m -v chl -v zooc -v o2 --minimum-latitude 60 --maximum-latitude 85 --minimum-longitude -10 --maximum-longitude 180 --dry-run --dataset-part originalGrid"
+    "**Standard Geographic Coordinates (Longitude/Latitude):**\n",
+    "\n",
+    "When working with conventional latitude/longitude projections, you specify the area using degree-based boundaries:\n",
```
gkoenig-mercator, line 1921: Not for this review, but we should think about when to use grid and projection. Because now they are nearly interchangeable in our minds but they are actually two different things, and picky readers or mathematically-inclined one could be confused by it.

doc/usage/stereo_subset.ipynb
```diff
@@ -2068,364 +1944,109 @@
...
-      "}\n"
-     ]
-    }
-   ],
    "source": [
-    "!copernicusmarine describe -i cmems_mod_arc_phy_my_hflux_P1M-m -e keywords,description,values,variables"
+    "Comprehensive Dataset Inspection:\n",
+    "\n",
+    "```bash\n",
+    "copernicusmarine describe -i cmems_mod_arc_phy_my_hflux_P1M-m -e keywords,description,values > describe_output.json\n",
```
gkoenig-mercator, line 2003: Maybe changing this command to "copernicusmarine describe -i cmems_mod_arc_phy_my_hflux_P1M-m -e units" as having the output hidden into a file could be confusing for novice users and I would assume non-novice users know about piping. Also we are interested in the units so I precised the units.

doc/usage/subset-usage.rst
```diff
@@ -42,26 +42,28 @@ Using log level DEBUG, a summary of the dataset subset is displayed. It is by de
...
 
-Sparse dataset can be subsetted using the ``subset`` command. The command will return the data in a tabular format i.e a pandas DataFrame, a CSV file or a Parquet database.
+Sparse datasets can be subset using the ``subset`` command, which returns the data in a tabular format, such as a Pandas DataFrame, a CSV file, or a Parquet database.
 
 **Example:**
 
 .. code-block:: bash
 
-  copernicusmarine subset -i cmems_obs-ins_arc_phybgcwav_mynrt_na_irr -y 45 -Y 90 -x -146.99 -X 180 -z 0 -Z 10 --start-datetime "2023-11-25T00:00:00" -T "2050-11-26T03:00:00" --dataset-part history -o todelete --platform-id B-Sulafjorden___MO --platform-id F-Vartdalsfjorden___MO --output-directory "some_place" --output-filename "my_subset"
+  copernicusmarine subset -i cmems_obs-ins_arc_phybgcwav_mynrt_na_irr -y 45 -Y 90 -x -146.99 -X 180 -z 0 -Z 10 --start-datetime "2023-11-25T00:00:00" -T "2050-11-26T03:00:00" --dataset-part history --platform-id B-Sulafjorden___MO --platform-id F-Vartdalsfjorden___MO
```
gkoenig-mercator, line 54: I propose using the version "copernicusmarine subset -i cmems_obs-ins_arc_phybgcwav_mynrt_na_irr -y 45 -Y 90 -x -146.99 -X 180 -z 0 -Z 10 --start-datetime "2023-11-25T00:00:00" -T "2050-11-26T03:00:00" --dataset-part history --platform-id B-Sulafjorden___MO --platform-id F-Vartdalsfjorden___MO" of the command, removing the end time. I understand your rationale for putting an end time far into the future. But I think it is confusing some some readers, who think this date in 2050 has some actual meaning.

doc/usage/stereo_subset.ipynb
elenaldm: I'd actually suggest something like:
"_The (longitude, latitude) projection is widely used in geospatial data. However, for regions near the poles - such as the Arctic and Antarctic - alternative coordinate systems, including the stereographic projection, are often employed. The specific projection applied to each dataset ..._"

doc/usage/stereo_subset.ipynb
elenaldm: I agree, maybe we can replace this point by something like:
"- It is more convenient for users who prefer to work directly with the coordinates of the original-grid"

doc/usage/stereo_subset.ipynb
gkoenig-mercator: I am unsure that longitude and latitude coordinates is actually a projection, but I think I overwent into details and you're right, we can definitely use this.

doc/usage/stereo_subset.ipynb
elenaldm: I agree, this point isn't "valid" here, as we are also providing the reprojection datacubes; I think we can simply skip it.

doc/usage/stereo_subset.ipynb
```diff
@@ -59,29 +62,32 @@
    "source": [
     "For a given dataset and version, the data in stereographic projection is in a different part.\n",
     "\n",
-    "This is why, when subsetting data based on the original-grid, it is necessary to explicitly specify the dataset part. In the command-line interface, this is achieved by using the option `--dataset-part originalGrid`. Similarly, in the Python interface, you would specify this parameter as follows: `dataset_part=\"originalGrid\"`."
+    "Consequently, when subsetting original-grid data, you must explicitly specify the dataset part:\n",
```
elenaldm, line 65: Uhmm what if something like (feel free to readapt):
"_Consequently, when downloading data in their original grid, you must explicitly specify the dataset part:
...
...
If missing, the toolbox will automatically return the default geographical coordinate system, in (longitude, latitude) projection_."

doc/usage/stereo_subset.ipynb
elenaldm: I agree. Do you know what is used by the User Support? We should follow the same definitions and terminologies everywhere

gkoenig-mercator: I will think a bit about this. For this version we can agree to be a bit vague and I will talk with the user support about it. I don't expect things to become crystal clear but at least we may come to an agreement.

doc/usage/subset-usage.rst
elenaldm: I'd not use "2050" as end date. In Copernicus Marine we don't have more than +10days of forecasts. Putting 2050 as end date gives "wrong expectations" to users. And also, the general definition of insitu observation data can't refer to the future, so it's also incorrect.

gkoenig-mercator: Ah yes I did not even think about the fact that it was not a forecast. It was probably just a typo for 2025 then

doc/usage/stereo_subset.ipynb
uriii3: Okey, I will delete it. (I understand you are referring to the one `"- It eliminates the requirement for reprojection or interpolation when converting to geographic coordinates \n",`.)

doc/usage/stereo_subset.ipynb
uriii3: Perfect!

doc/usage/stereo_subset.ipynb
uriii3: I'll resolve it for now!

doc/usage/stereo_subset.ipynb
uriii3: Mmmh actually the `-e` part is to exclude the fields! But I get the idea, I will change the output and try to add dataset label and the units only.

uriii3: I end up using: `copernicusmarine describe -i cmems_mod_arc_phy_my_hflux_P1M-m -r coordinates,label,name,service_name -e values`, I think it is indeed more explanatory! Thanks for the feedback.

PR #317, opened 2025-04-18
fix: geo input to 0 would not be selected

renaudjester:
Fixes a bug where setting `maximum_longitude` (for eg) to 0 would result into selecting the whole dataset range

example: 
dataset_insitu = "cmems_obs-ins_arc_phybgcwav_mynrt_na_irr"

response = copernicusmarine.subset(
      dataset_id=dataset_insitu,
      minimum_latitude=45,
      maximum_latitude=90,
      minimum_longitude=-45,
      maximum_longitude=0,
      minimum_depth=0,
      maximum_depth=10,
      variables=["TEMP"],
      start_datetime="2023-11-25T00:00:00",
      end_datetime="2024-11-26T03:00:00",
      dataset_part="history"
)


uriii3 (approved):
the only critic is that the title doesn't not match the updates in the code, no? I love the restructuring of the function you made, thanks! It looks indeed cleaner now!

Theophile-Varnier (approved):

PR #327, opened 2025-05-08
fix: add product doi as column and product doi at dataset metadata level

renaudjester:
Fix [CMT-230](https://cms-change.atlassian.net/browse/CMT-230)

----
📚 Documentation preview 📚: https://copernicusmarine--327.org.readthedocs.build/en/327/

[CMT-230]: https://cms-change.atlassian.net/browse/CMT-230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/copernicusmarine/catalogue_parser/catalogue_parser.py
+++ b/copernicusmarine/catalogue_parser/catalogue_parser.py
@@ -63,13 +63,18 @@ def get_dataset_metadata(
         ).get(dataset_id)
         if not product_ids:
             raise DatasetNotFound(dataset_id)
-        url_dataset_jsons_mapping: dict[str, dict] = {}
+        url_dataset_jsons_mapping: dict[str, tuple[dict, Optional[str]]] = {}
         for product_id in product_ids.split(","):
             url = f"{stac_url}/{product_id}/product.stac.json"
             product_json = connection.get_json_file(url)
             product_collection = pystac.Collection.from_dict(product_json)
             product_datasets_metadata_links = (
                 product_collection.get_item_links()
+            )
+            digital_object_identifier = (
+                product_collection.extra_fields.get("sci:doi", None)
+                if product_collection.extra_fields
+                else None
             )
             datasets_metadata_links = [
                 dataset_metadata_link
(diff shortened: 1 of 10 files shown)


copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -63,14 +63,19 @@ def get_dataset_metadata(
         ).get(dataset_id)
         if not product_ids:
             raise DatasetNotFound(dataset_id)
-        url_dataset_jsons_mapping: dict[str, dict] = {}
+        url_dataset_jsons_mapping: dict[str, tuple[dict, Optional[str]]] = {}
```
Theophile-Varnier, line 66: Any chance we could create a class for this?

renaudjester: I agree to do classes for complicated typings, but I would say that it should be for something that we reuse several times (so that it makes it clearer, not more complex) 

I will take a look to see if I can make it work :D

Theophile-Varnier: Fair enough

copernicusmarine/catalogue_parser/models.py
```diff
@@ -591,6 +591,9 @@ class CopernicusMarineDataset(BaseModel):
     dataset_id: str
     #: The dataset name.
     dataset_name: str
+    #: Digital object identifier or doi from
+    #: the product the dataset belongs to.
+    digital_object_identifier: Optional[str]
```
uriii3, line 596: It is a little weird for me that if the DOI is in the dataset level, no? Just makes sense in terms of simplifying code, but maybe it is worth it too to have it like this!

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -360,6 +372,12 @@ def _transform_dataframe(
         )
     )
 
+    df["product_doi"] = (
+        f"https://doi.org/{product_doi}"
+        if product_doi and "https://" not in product_doi
+        else product_doi
+    )
```
uriii3, line 379: `if not product_doi`, should we add a column with a 'None', anyway? It is a little weird, no?

doc/usage/subset-usage.rst
```diff
@@ -97,7 +97,8 @@ The output will contain the following columns:
 - ``value``: The measurement value.
 - ``value_qc``: The quality control indicator of the value.
 - ``institution``: The institution that produced the data and is affiliated with the platform.
-- ``doi``: The DOI of the data.
+- ``doi``: The DOI of the data, from the data producers.
+- ``product_doi``: The DOI of the product.
```
uriii3, line 101: Here I would try to explain a bit better the difference between both. Something like "`The DOI of the product inside the copernicus marine catalogue.`" (?) I think you have better ideas with the writing but maybe a little explanation on why the differences help understand better both parameters.

copernicusmarine/download_functions/download_sparse.py
renaudjester: Yep :D we want the columns to be always present even if None

copernicusmarine/catalogue_parser/models.py
renaudjester: Yes I know but if I want to get it it's a lot of code modification so I thought I can afford to duplicate it

uriii3: Perfect, looks good to me!

uriii3 (approved):

doc/usage/subset-usage.rst
renaudjester: @elenaldm any idea?

Theophile-Varnier (approved):

doc/usage/subset-usage.rst
elenaldm: 🤔 what about something like:
`doi`: The DOI of the data, based on the originating institution.
`product_doi`: Product's DOI in the Copernicus Marine Catalog.
?

renaudjester: Perfect!

uriii3 (approved):

PR #329, opened 2025-05-12
fix: return empty dataframe if no data

renaudjester:
fix [CMT-233](https://cms-change.atlassian.net/browse/CMT-233)

----
📚 Documentation preview 📚: https://copernicusmarine--329.org.readthedocs.build/en/329/

[CMT-233]: https://cms-change.atlassian.net/browse/CMT-233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/copernicusmarine/download_functions/download_sparse.py
+++ b/copernicusmarine/download_functions/download_sparse.py
@@ -228,7 +228,11 @@ def _read_dataframe_sparse(
         df = _transform_dataframe(
             df, subset_request.vertical_axis, platforms_metadata
         )
-
+    if df.empty:
+        logger.info(
+            "No data found for the given parameters. "
+            "Please check your request and try again."
+        )
     return (
         df,
         variables,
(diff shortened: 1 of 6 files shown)


copernicusmarine/download_functions/download_sparse.py
```diff
@@ -342,6 +342,8 @@ def _transform_dataframe(
     Transform the dataframe to match the expected format to be consistent with MyOceanPro
     and Copernicus Marine Services.
     """  # noqa
+    if df.empty:
+        return df
```
uriii3, line 346: wouldn't you like some sort of warning, here? something like:
`logger.warning("There is no data in the described subset region")` that would improve the users understanding of why the data frame is empty?

uriii3 (approved):
Up to you, I guess it makes sense to just return it as it is and maybe multiple reasons for that to happen.

copernicusmarine/download_functions/download_sparse.py
renaudjester: Good idea! not for a warning but indeed at least a debug

PR #335, opened 2025-05-14
doc: improve binaries doc examples

renaudjester:
For the examples with binaries to be more user friendly and human readable

----
📚 Documentation preview 📚: https://copernicusmarine--335.org.readthedocs.build/en/335/


uriii3 (approved):
Okay! Have you checked if they work?

PR #349, opened 2025-05-28
[CMT-196] Run tests with poetry

uriii3:
As a first iteration to improve the tests, we want to setup the tests with poetry, so that across all platforms the implementation will be easier and more unified.

----
📚 Documentation preview 📚: https://copernicusmarine--349.org.readthedocs.build/en/349/


uriii3:
Maybe in the future we can revisit this issue here https://github.com/pytest-dev/pytest-xdist/issues/385, as it looks like if we can serialise. (and insulate) the specific 'fast' test, we would ensure that it still goes fast :)

.github/workflows/tests.yml
```diff
@@ -18,20 +18,33 @@ jobs:
...
+        run: poetry install --no-interaction --with test
       - name: Run tests
         env:
           COPERNICUSMARINE_SERVICE_USERNAME: ${{ secrets.COPERNICUSMARINE_SERVICE_USERNAME }}
           COPERNICUSMARINE_SERVICE_PASSWORD: ${{ secrets.COPERNICUSMARINE_SERVICE_PASSWORD }}
-        run: make run-tests
-        shell: micromamba-shell {0}
+        run: |
+          source $VENV
+          pytest tests -n auto --dist=loadgroup --verbose -vv --durations=0 --log-cli-level=info --basetemp="tests/downloads" --junitxml=report.xml --log-format "%(asctime)s %(levelname)s %(message)s" --log-date-format "%Y-%m-%d %H:%M:%S"
```
renaudjester, line 50: In the makefile :D

.github/workflows/tests.yml
```diff
@@ -18,20 +18,33 @@ jobs:
...
         with:
-          micromamba-version: '1.5.6-0'
-          micromamba-binary-path: ${{ runner.temp }}/bin/micromamba
-          environment-file: conda_environment_test.yaml
-          environment-name: copernicusmarine_test
-          condarc-file: .condarc
-          cache-environment: true
-          post-cleanup: 'all'
-
+            python-version: '3.12'
```
renaudjester, line 25: I think we were in 3.13

.github/workflows/tests.yml
```diff
@@ -18,20 +18,33 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+      - name: Install dependencies
+        run: poetry install --no-interaction --no-root
+      - name: Install ncdump dependency
+        run: sudo apt-get install netcdf-bin -y
+      - name: Install library
+        run: poetry install --no-interaction --with test
```
renaudjester, line 43: Why the two installs for poetry? Can it be done in one?

Makefile
```diff
@@ -39,7 +39,6 @@ check-format:
 
 run-tests: SELECTED_ENVIRONMENT_NAME = ${TEST_ENVIRONMENT_NAME}
 run-tests:
-	${ACTIVATE_ENVIRONMENT}
```
renaudjester, line 42: ❤️

pyproject.toml
```diff
@@ -43,6 +43,13 @@ syrupy = "^4.8.1"
 numpydoc = "^1.8.0"
 netcdf4 = "^1.7.2"
 
+[tool.poetry.group.test]
+optional = true
+
+[tool.poetry.group.test.dependencies]
+pytest-xdist = "^3.7.0"
+compliance-checker = { version = "^5.1.1", python = ">3.10" }
```
renaudjester, line 51: don't you also need syrupy and others? basically everything that is in the `conda_environment_test.yaml`? And do you need this conda yaml in the end?

tests/test_command_line_interface.py
```diff
@@ -1103,6 +1104,7 @@ def test_log_level_debug(self, tmp_path):
             tmp_path / output_filename, response
         )
 
+    @pytest.mark.xdist_group(name="sequential")
```
renaudjester, line 1107: what is this for? Does it help? If yes why do you need to increase the timeout anyway?

uriii3: It was meant to help with this: https://github.com/pytest-dev/pytest-xdist/issues/385 but it doesn't look like. I can just delete it and write a TODO, with the issue, as a reminder that when it is solved we could check it again?

.github/workflows/tests.yml
renaudjester: Why do you need to pass it in the make file? 

you can do: 

```
run: |
          source $VENV
          make tests
```

Also normally with actions you pass the env var in the env section: 

```
env: VENV={{value}}
run: |
          make tests
```

tests/test_command_line_interface.py
uriii3: Okey, a small update: I removed it and it crashed with the new timeout. So it looks like indeed the code works a bit, but not as the level that it was before. I will maintain the line and then we can recheck to have the old timeout!

pyproject.toml
uriii3: mmh well, I don't think so! One thing is the libraries that you need to develop (syrupy to update snapshots, precommit to commit, even the types...) but another set is the ones you need to test! and we don't need syrupy to run the tests, as they have not raised any error! So I think it is best to maintain them separately, with the group dev and test accordingly!

.github/workflows/tests.yml
```diff
@@ -18,20 +18,32 @@ jobs:
...
         with:
-          micromamba-version: '1.5.6-0'
-          micromamba-binary-path: ${{ runner.temp }}/bin/micromamba
-          environment-file: conda_environment_test.yaml
-          environment-name: copernicusmarine_test
-          condarc-file: .condarc
-          cache-environment: true
-          post-cleanup: 'all'
-
+            python-version: '3.12'
```
renaudjester, line 26: I thought we were using python 3.13?

pyproject.toml
renaudjester: I am pretty sure that the "dev" env is being install by default because I don't see how your tests would pass without the dependencies in the dev env

If it's the case then we can group them

.github/workflows/tests.yml
renaudjester: It works here: https://github.com/mercator-ocean/copernicus-marine-toolbox/blob/main/.github/workflows/dependencies-tests.yml

.github/workflows/tests.yml
```diff
@@ -12,26 +12,48 @@ jobs:
...
-          condarc-file: .condarc
-          cache-environment: true
-          post-cleanup: 'all'
+            python-version: '3.13'
+      - name: Check Python version
+        run: |
+          python --version
+          python3.13 --version
+          which python
+          which python3.13
```
renaudjester, line 33: Can be deleted

.github/workflows/tests.yml
```diff
@@ -12,26 +12,48 @@ jobs:
...
+      - name: Install and configure Poetry
+        run: |
+          python3.13 -m pip install poetry
+          which poetry
```
renaudjester, line 37: can be deleted

.github/workflows/tests.yml
```diff
@@ -12,26 +12,48 @@ jobs:
...
+      - name: Install library
+        run: python3.13 -m poetry install --no-interaction --with dev
       - name: Run tests
         env:
           COPERNICUSMARINE_SERVICE_USERNAME: ${{ secrets.COPERNICUSMARINE_SERVICE_USERNAME }}
           COPERNICUSMARINE_SERVICE_PASSWORD: ${{ secrets.COPERNICUSMARINE_SERVICE_PASSWORD }}
-        run: make run-tests
-        shell: micromamba-shell {0}
+        run: |
+          source .venv/bin/activate
```
renaudjester, line 58: Not gonna work on windows but you can deal with that later

uriii3: hehehe exactly that is what I thought heheh

PR #358, opened 2025-06-18
Update coverage

uriii3:
Just change the place where the coverage badge is in, but most importantly rerun the tests to update them in the correct place.

----
📚 Documentation preview 📚: https://copernicusmarine--358.org.readthedocs.build/en/358/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,11 @@ on:
       - "main"
       - "release/**"
       - "pre-releases/**"
+  push:
+    branches:
+      - "main"
+      - "release/**"
+      - "pre-releases/**"
 
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -6,6 +6,11 @@ on:
       - "main"
       - "release/**"
       - "pre-releases/**"
+  push:
```
renaudjester, line 9: You need to select only the one with coverage if possible

.github/workflows/tests.yml
```diff
@@ -6,6 +6,11 @@ on:
...
+    branches:
+      - "main"
+      - "release/**"
+      - "pre-releases/**"
```
renaudjester, line 13: not needed for release and pre-releases?

renaudjester (approved):

PR #370, opened 2025-07-02
fix: read only open zarr

renaudjester:
fix [issue reported here ](https://github.com/zarr-developers/zarr-python/issues/3189)

----
📚 Documentation preview 📚: https://copernicusmarine--370.org.readthedocs.build/en/370/


veenstrajelmer:
Thanks for fixing this so quickly. And I see a release with temporary zarr upper limit, perfect, very helpful!

PR #380, opened 2025-07-09
[doc-update] remove old auth system mention

uriii3:
Remove the mentioning of the old authentication system from the error.

Do you think that we should say something, between now and when we remove the old-car completely from the code, about checking that the credentials are okey? Or would it be a problem onwards, once solved?

Let me know what you think!

----
📚 Documentation preview 📚: https://copernicusmarine--380.org.readthedocs.build/en/380/

### DIFF
--- a/doc/usage/login-usage.rst
+++ b/doc/usage/login-usage.rst
@@ -17,16 +17,10 @@ The ``login`` command creates a configuration file called ``.copernicusmarine-cr
 
 If the ``.copernicusmarine-credentials`` file already exists, the system will ask for confirmation before overwriting it. You can also use option ``–-force-overwrite`` to skip confirmation.
 
-New Copernius Marine authentication system
+Old Copernius Marine authentication system
 -------------------------------------------
 
-A new Copernius Marine authentication system will be released in the following months after the release of the Copernicus Marine Toolbox version 2.0.0.
-From 2.0.0, the toolbox should be able to handle both the old and the new authentication systems.
-
-If you are blocking some domains, you will need to authorize the domain ``auth.marine.copernicus.eu`` to be able to connect when the old system is decomissioned.
-
-.. note::
-    One of limitation of the old system is that it goes through HTTP (through redirections) and not HTTPS. The new system will use HTTPS only.
+From 9th July 2025, the old Copernicus Marine authentication system is deprecated.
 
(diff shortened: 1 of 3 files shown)


gkoenig-mercator:
And in [this paragraph](https://toolbox-docs.marine.copernicus.eu/en/v2.1.0/usage/login-usage.html#access-points-migration-and-evolution), I have the impressions that some things have been removed. If they have, we should maybe update the documentation:

"The use of motuclient file is deprecated and will be removed in the future"

"Also, the hosts nrt.cmems-du.eu and my.cmems-du.eu are deprecated for ‘netrc’ files and will be removed in the future" 

If this is the case, we could replace them with:

"The use of motuclient file is deprecated and has been removed the 09/07/2025" or "version 2.2.0"

"Also, the hosts nrt.cmems-du.eu and my.cmems-du.eu are deprecated for ‘netrc’ files and have been removed in version 2.2.0" 

thanks for the answer :)

doc/usage/login-usage.rst
```diff
@@ -17,16 +17,10 @@ The ``login`` command creates a configuration file called ``.copernicusmarine-cr
 
 If the ``.copernicusmarine-credentials`` file already exists, the system will ask for confirmation before overwriting it. You can also use option ``–-force-overwrite`` to skip confirmation.
 
-New Copernius Marine authentication system
+Old Copernius Marine authentication system
 -------------------------------------------
```
renaudjester, line 21: I think we can just delete this paragraph IMO

uriii3: I think so too! I wanted to have your view and propose something in positive, but I think I would delete it too!

renaudjester: if we keep it, which I think is not really the role of the documentation now (more an article on the help center) then I would name it: "``Former Copernius Marine authentication``" 

and then add some context: 
Previously you could authenticate through this URL, but not anymore, it has been decommissioned. 

Important that you don't say it's "deprecated", it is not, it is deleted/decommissioned/not available anymore

renaudjester: But this is more of a migration information that in the next release will be obsolete and need to be deleted

renaudjester (approved):

PR #381, opened 2025-07-10
CMT-210 delete old auth related code

renaudjester:
----
📚 Documentation preview 📚: https://copernicusmarine--381.org.readthedocs.build/en/381/


doc/changelog/changelog.rst
uriii3, line 1: why don't we have here the v213? should we add it here? (I just saw your comments too about it!)

doc/changelog/v221.rst
```diff
@@ -0,0 +1,12 @@
...
+
+Patch release with bug fixes and improvements.
+
+Login
+------
+
+Fixes
+^^^^^
+
+* Deleted the code from the old authentication system. It fixes some issues where the error returned by the Toolbox was not the proper ones.
```
uriii3, line 12:
```suggestion
* Deleted the code from the old authentication system. It fixes some issues where the error returned by the Toolbox was not the proper one.
```
To make sense semantically, no?

tests/test_login.py
```diff
@@ -474,3 +447,18 @@ def test_get_data_with_motuclient_file(self, tmp_path):
...
+        responses.add(
+            responses.POST,
+            COPERNICUS_MARINE_AUTH_SYSTEM_TOKEN_ENDPOINT,
+            json=None,
+            status=503,
+        )
+
+        try:
+            login(username="toto", password="lololo")
+            assert True, "should raise before"
```
uriii3, line 462: don't we do an `assert False` for this sort of catching the raising error?

doc/changelog/changelog.rst
renaudjester: Let's do another PR for this not to mix everything up

PR #386, opened 2025-07-15
[chore] [CMT-278] Describe function more resilient

uriii3:
Trying to make the describe function more resilient, so that if a part of a dataset is not compliant or has any problem, it does not break the whole parsing of the describe.

----
📚 Documentation preview 📚: https://copernicusmarine--386.org.readthedocs.build/en/386/


copernicusmarine/catalogue_parser/models.py
```diff
@@ -652,15 +655,38 @@ def sort_versions(self) -> None:
...
+                    dataset_item.url,
+                )
+            except Exception as e:
+                if stop_at_failure:
+                    logger.error(
+                        f"Failed to parse part {dataset_item.parsed_part} "
+                        f"for dataset {dataset_item.parsed_id}.",
+                    )
+                    logger.error(
+                        "Stopping describe. "
```
renaudjester, line 676: Could it be that you are stopping it when doing a get for eg? If not, maybe this could be interesting too? Since you do it at part level

copernicusmarine/catalogue_parser/models.py
```diff
@@ -652,15 +655,38 @@ def sort_versions(self) -> None:
...
     ) -> None:
         all_versions = set()
         for dataset_item in dataset_items:
             dataset_version = dataset_item.parsed_version
-            part = CopernicusMarinePart.from_metadata_item(
-                dataset_item.stac_item,
-                dataset_item.parsed_part,
-                dataset_item.url,
-            )
+            try:
```
renaudjester, line 663: Could it fail at product level too?

copernicusmarine/core_functions/documentation_utils.py
```diff
@@ -124,6 +124,10 @@
         "Maximum number of concurrent requests (>=1). Default 15. The command uses "
         "a thread pool executor to manage concurrent requests."
     ),
+    "STOP_AT_FAILURE_HELP": (
+        "If set to True, the function will stop at the first failure encountered "
+        "during the parsing of the catalogue."
```
renaudjester, line 129: fetching too right?

tests/test_describe.py
```diff
@@ -80,6 +85,34 @@ def test_describe_exclude_datasets(self):
...
+                assert False, "Expected an exception to be raised"
+            except Exception:
+                assert True
+                assert "Stopping describe" in caplog.text
+
+    @mock.patch(
+        "requests.Session.get",
+        side_effect=mocked_stac_requests_get,
+    )
+    def test_describe_without_stop_at_failure(self, argument, caplog):
```
renaudjester, line 107: you should also test if a fetch fails ie on of the URL you try to get fails

copernicusmarine/catalogue_parser/models.py
renaudjester: No what I mean is that: let's say there is an error in the product STAC file, would it fail the whole describe to try to fetch or parse the product.stac.json?

copernicusmarine/core_functions/documentation_utils.py
renaudjester: Also just calling GET on a dataset.stac.json might fail, so maybe we need to make sure that connection errors doesn't result in full stop of the process

copernicusmarine/catalogue_parser/models.py
uriii3: Yes, I think so! Do we want to make it also resilient in that sense? Also, what sort of errors do you imagine to have on stac? I don't know if we do that many checks regarding their compliancy.

renaudjester: But it's exactly the same as the datasets: what you are doing here is being sure that we don't crash the whole process if there is an error in this: 

- fetch dataset-version-part (requests.get(.....dataset.stac.json))
- parse the json and get the data into a CopernicusMarineDataset

Well the exact same problems can arise from this: 

- fetch product (requests.get(.....product.stac.json))
- parse the json and get the data into a CopernicusMarineProduct

uriii3: Looks clear to me. I will look at it hehe 🙌 (the fetching part is also done now! :) )

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -284,7 +305,16 @@ def fetch_dataset_items(
...
+            item_json = connection.get_json_file(url)
+        except Exception as e:
+            logger.debug(
+                f"Failed to fetch or parse JSON for dataset URL: {url}"
+            )
+            if stop_at_failure:
+                raise e
+            log_exception_debug(e)
+            continue
         item = _parse_dataset_json_to_pystac_item(item_json)
```
uriii3, line 318: I don't know if I have to consider the errors of it not being a pystac item... I did it with product but I'm unsure. What do you think of it?

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -28,7 +31,9 @@
 
 
 def get_dataset_metadata(
-    dataset_id: str, marine_datastore_config: MarineDataStoreConfig
+    dataset_id: str,
+    marine_datastore_config: MarineDataStoreConfig,
+    stop_at_failure: bool = False,
```
renaudjester, line 36: Maybe it doesn't need to be a kwarg if you set it everywhere anyway 🤔

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -419,21 +467,21 @@ def parse_catalogue(
...
-        )
-    ]
+    products_metadata: list[CopernicusMarineProduct] = []
+    for product_item in marine_data_store_root_collections:
+        if product_item:
+            if product_metadata := _construct_marine_data_store_product(
+                product_item, stop_at_failure
+            ):
+                product_metadata.datasets
+                products_metadata.append(product_metadata)
```
renaudjester, line 481: Why this one? wasn't clear enough the comprehension list?

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -419,21 +467,21 @@ def parse_catalogue(
...
-            ).datasets
-        )
-    ]
+    products_metadata: list[CopernicusMarineProduct] = []
+    for product_item in marine_data_store_root_collections:
+        if product_item:
+            if product_metadata := _construct_marine_data_store_product(
+                product_item, stop_at_failure
+            ):
+                product_metadata.datasets
```
renaudjester, line 480: what is it for?

tests/resources/mock_stac_catalog_WAW3/mock_dataset_irrelevant.py
renaudjester, line 1: why is it "irrelevant"?

tests/test_describe.py
```diff
@@ -80,6 +85,106 @@ def test_describe_exclude_datasets(self):
...
+                show_all_versions=True,
+            )
+            assert "Failed to parse part" in caplog.text
+            assert "Skipping part." in caplog.text
+            assert (
+                "Failed to fetch or parse JSON for product URL:" in caplog.text
+            )
+            assert (
+                "Failed to fetch or parse JSON for dataset URL:" in caplog.text
+            )
```
renaudjester, line 184: maybe check that the URL is the right one

tests/test_describe.py
```diff
@@ -80,6 +85,106 @@ def test_describe_exclude_datasets(self):
...
+    ):
+        with caplog.at_level(logging.DEBUG, logger="copernicusmarine"):
+            try:
+                describe(
+                    stop_at_failure=True,
+                    product_id="PRODUCT_W_ERRORS",
+                )
+                assert False, "Expected an exception to be raised"
+            except Exception:
+                assert True
```
renaudjester, line 164: assert True is not useful right?

copernicusmarine/core_functions/describe.py
```diff
@@ -26,6 +26,7 @@ def describe_function(
     max_concurrent_requests: int,
     disable_progress_bar: bool,
     staging: bool,
+    stop_at_failure: bool,
```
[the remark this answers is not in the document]
uriii3: Mmmh I'm just following the specifications! We can open a little discussion if needed, I will now.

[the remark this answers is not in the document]
uriii3: https://cms-change.atlassian.net/browse/CMT-278?focusedCommentId=68516

copernicusmarine/catalogue_parser/catalogue_parser.py
uriii3: Mmh yes but then I was having a lot of trouble if the retired value was 'None'. This way, I can differentiate the case in which there exists a returned `product_metadata` and when that one is None.

copernicusmarine/catalogue_parser/catalogue_parser.py
uriii3: It was there before, and I was using it with a try so it made sense. It does not anymore now, though.

copernicusmarine/catalogue_parser/catalogue_parser.py
uriii3: Committed

copernicusmarine/catalogue_parser/catalogue_parser.py
renaudjester: Well you would do it in the comprehension list but it's fine like this

Though before there was a check that `product_metadata.datasets` was not empty, and it's not here anymore

doc/changelog/v230.rst
```diff
@@ -17,3 +18,11 @@ Fixes
...
 
 * Fixed an issue where the ``valid_range`` attribute for some variables of some netcdf files was not in the correct format and hence not CF-compliant. The Toolbox now checks the format of the ``valid_range`` attribute and converts it to the same type as the data.
+
+Describe
+----------
+
+New-Features
+^^^^^^^^^^^^^^^
+
+* Add a flag `raise_on_error` to the `describe` command to raise an error when the fetching or parsing fails. Default is set to ``False``.
```
renaudjester, line 28: add link to the new doc you created

doc/usage/describe-usage.rst
```diff
@@ -236,3 +236,9 @@ See `the raise-if-updating <raise-if-updating>`_ option to be sure your requeste
 .. warning::
 
     ``arco_updating_start_date`` is a date within the dataset while ``arco_updated_date`` is a real-world timestamp.
+
+``raise_on_error`` option
+----------------------------
+The ``describe`` command has a ``--raise-on-error`` option that allows you to raise an error if the fetching or parsing of the metadata fails.
```
renaudjester, line 242: Explain at which level it would fail so that users can see why they should or should not use it?

gkoenig-mercator: I thought we could explicit the default behavior with this:

 "Add a stop_at_failure flag to the describe command. If set to True, it raises an error on any parsing or metadata compliance failure. By default (False), it skips errors and returns only valid products.""

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -419,21 +467,20 @@ def parse_catalogue(
...
-            ).datasets
-        )
-    ]
+    products_metadata: list[CopernicusMarineProduct] = []
+    for product_item in marine_data_store_root_collections:
+        if product_item:
+            if product_metadata := _construct_marine_data_store_product(
+                product_item, raise_on_error
+            ):
+                products_metadata.append(product_metadata)
```
renaudjester, line 480: You are not checking if `product_metadata.dataset` is empty

I don't remember how important it is, but it was done before

copernicusmarine/command_line_interface/group_describe.py
```diff
@@ -120,6 +127,7 @@ def describe(
     disable_progress_bar: bool,
     log_level: str,
     staging: bool,
+    raise_on_error: bool = False,
```
renaudjester, line 130:
```suggestion
    raise_on_error: bool,
```

copernicusmarine/core_functions/documentation_utils.py
```diff
@@ -124,6 +124,10 @@
         "Maximum number of concurrent requests (>=1). Default 15. The command uses "
         "a thread pool executor to manage concurrent requests."
     ),
+    "RAISE_ON_ERROR_HELP": (
+        "If set to True, the function will stop at the first failure encountered "
+        "during the parsing and fetching of the catalogue."
+    ),
```
renaudjester, line 130:
```suggestion
    "RAISE_ON_ERROR_HELP": (
        "If set to True, the function will raise at the first error encountered "
        "during the parsing and fetching of the catalogue. Default, False."
    ),
```

doc/changelog/v230.rst
```diff
@@ -9,6 +9,7 @@ General
 New features
 ^^^^^^^^^^^^^^^
 
+* Increased the resilience when parsing or fetching the metada. Make the describe more robust and consequently the `subset` and `get` commands as well.
```
renaudjester, line 12:
```suggestion
* Increased the resilience when parsing or fetching the metadata. Make the ``describe`` more robust and consequently the ``subset`` and ``get`` commands as well.
```

doc/usage/describe-usage.rst
```diff
@@ -236,3 +236,12 @@ See `the raise-if-updating <raise-if-updating>`_ option to be sure your requeste
 .. warning::
 
     ``arco_updating_start_date`` is a date within the dataset while ``arco_updated_date`` is a real-world timestamp.
+
+
+.. _raise_on_error:
+
+``raise_on_error`` option
+----------------------------
+A flag that if set to True raises an error on any parsing or metadata compliance failure. By default (False), it skips errors and returns only valid products and datasets.
```
renaudjester, line 245:
```suggestion
A flag that if set to ``True`` raises an error on any parsing or metadata compliance failure. By default (``False``), it skips errors and returns only valid products and datasets.

Set it to ``True`` if you want to be sure of the integrity of the result. Otherwise, setting it to ``False`` will increase the availability and still return you the result of the ``describe`` even in case of a momentary problem with a dataset or product. 

.. note: 

    Results are skipped at the part level as the lowest level. It could be that a dataset is correctly parsed, but might be missing a part that couldn't be fetched or parsed. A full dataset would be erroneous or even a full product. Please check to debug logs to know more. 

```

copernicusmarine/catalogue_parser/catalogue_parser.py
uriii3: I guess we can add a check!

renaudjester (approved):

PR #399, opened 2025-08-20
End-Of-Life Page

uriii3:
Let's add and EOL page to the documentation... with all the minor versions there it would need to be updated at every release, so let's talk it, but I think the information might be relevant. Also, all the changes from version to version should be included in the changelog... let's see.

### DIFF
--- a/doc/end-of-life.rst
+++ b/doc/end-of-life.rst
@@ -1,0 +1,73 @@
+Copernicus Marine Toolbox – Versioning and End-of-Life (EOL) Policy
+=======================================================================
+
+Purpose
+-------
+
+This document defines the versioning scheme and end-of-life (EOL) timelines for the **Copernicus Marine Toolbox** (CLI and Python package). It helps users plan upgrades and understand support timelines.
+
+----
+
+Versioning Scheme
+-----------------
+
+The Copernicus Marine Toolbox follows **semantic versioning**:
+
+MAJOR.MINOR.PATCH
+^^^^^^^^^^^^^^^^^
+
+- **MAJOR** – Breaking changes (incompatible APIs, major architectural changes)
+- **MINOR** – Backward-compatible features and improvements
+- **PATCH** – Backward-compatible bug fixes
+
+Example: ``2.1.4`` → major=2, minor=1, patch=4.
+
+----
+
+Support Policy
+--------------
+
+- Only the **latest minor/patch** in a major branch is supported.
+- After EOL, no further fixes, updates, or security patches will be released.
+
+The v1 is **no longer supported** and should be replaced with the ``v2``.
+
+----
+
+Release and End-of-Life Timeline
+---------------------------------
+
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+| Version             | Release Date | End of Support | Latest version    |Notes                                      |
++=====================+==============+================+===================+===========================================+
+| 1.x                 | 2023-06      | 2025-07        | :del:`1.3.6`         | :del:`First public release of new toolbox`       |
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+| 2.0                 | 2025-01      | TBD            | 2.0.1             | Major release with new features and improvements |
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+| 2.1                 | 2025-05      | TBD            | 2.1.3             | New important features                    |
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+| 2.2                 | 2025-07      | TBD            | 2.2.1             | Current stable version                    |
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+| 2.3                 | 2025-09      | TBD            | Not yet released  | Coming soon
++---------------------+--------------+----------------+-------------------+-------------------------------------------+
+
+
+.. note::
+   Dates are indicative. For the latest updates, visit the `Copernicus Marine User Support Portal <https://marine.copernicus.eu>`_.
+
+
+Dependencies and Versions
+----------------------------
+Although for a more thorough view of the dependencies versions allowed you can check the documentation of each version, here is a summary of the dependencies versions for some of the more relevant libraries used:
+
+========  ========  ===========  ===========
+Version   Python    Xarray       Zarr
+========  ========  ===========  ===========
+2.0       3.9+      2023.4.0+    2.13.3+
+2.1       3.9+      2023.4.0+    2.13.3 to 3.0.8
+2.2       3.9+      2023.4.0+    2.13.3+
+========  ========  ===========  ===========
+
+.. note::
+   For complete dependency specifications including optional packages,
+   see the ``requirements.txt`` file for each version.


doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
...
+Dependencies and Versions
+----------------------------
+Although for a more thorough view of the dependency versions allowed you can check the documentation of each version, here is a summary of the dependency versions for some of the more relevant libraries used:
+
+========  ========  ===========  ===========
+Version   Python    Xarray       Zarr
+========  ========  ===========  ===========
+2.0       3.9+      2023.4.0+    2.13.3+
+2.1       3.9+      2023.4.0+    2.13.3 to 3.0.8
+2.2       3.9+      2023.4.0+    2.13.3+
```
renaudjester, line 69: Maybe a little sentence explaining that when you say 2.0 you mean: the latest patch of 2.0? Of name the column Minor Version?

doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
...
+
+
+Dependencies and Versions
+----------------------------
+Although for a more thorough view of the dependency versions allowed you can check the documentation of each version, here is a summary of the dependency versions for some of the more relevant libraries used:
+
+========  ========  ===========  ===========
+Version   Python    Xarray       Zarr
+========  ========  ===========  ===========
+2.0       3.9+      2023.4.0+    2.13.3+
```
renaudjester, line 67: I think prefer the classic `>=` than `+` 

for eg: `>=2.13.3, <=3.0.8` instead of `2.13.3 to 3.0.8`

doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
...
+.. note::
+   Dates are indicative. For the latest updates, visit the `Copernicus Marine User Support Portal <https://marine.copernicus.eu>`_.
+
+
+Dependencies and Versions
+----------------------------
+Although for a more thorough view of the dependency versions allowed you can check the documentation of each version, here is a summary of the dependency versions for some of the more relevant libraries used:
+
+========  ========  ===========  ===========
+Version   Python    Xarray       Zarr
```
renaudjester, line 65:
```suggestion
Copernicusmarine   Python    Xarray       Zarr
```

?

doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
+=======================================================================
+Versioning and End-of-Life (EOL) Policy
+=======================================================================
+
```
renaudjester, line 4: What about a section at the beginning like: "Important information" or "News" or something like this with the most important info like: "v2.3 will not support python 3.10" @gkoenig-mercator

doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
...
+Purpose
+-------
+
+This document defines the versioning scheme and end-of-life (EOL) timelines for the **Copernicus Marine Toolbox** (CLI and Python package). It helps users plan upgrades and understand support timelines.
```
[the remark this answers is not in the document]
gkoenig-mercator: If it is the same timeline for both the Python version and its CLI wrapper, maybe we can have "timeline" in singular form?

doc/end-of-life.rst
gkoenig-mercator: The little sentence could go in the paragraph at the beginning that would take the form :"For a more detailed overview of the allowed dependency versions, you can refer to the documentation for each version of the Copernicus Marine Toolbox. The version listed here always corresponds to the latest patch. Below is a summary of the dependency versions for some of the most relevant libraries used: "

I don't know if we need to indicate the patch version as well though, or indicate that it will the one automatically retrieve with any installer.

doc/end-of-life.rst
gkoenig-mercator: The user support also say that they tend to have issues reported to them with dask and would like it to be in this table as well.

doc/end-of-life.rst
```diff
@@ -0,0 +1,74 @@
...
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
+| 2.0                 | 2025-01      | TBD            | 2.0.1             | Major release with new features and improvements     |
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
+| 2.1                 | 2025-05      | TBD            | 2.1.3             | New important features                               |
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
+| 2.2                 | 2025-07      | TBD            | 2.2.1             | Current stable version                               |
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
+| 2.3                 | 2025-09      | TBD            | Not yet released  | Coming soon                                          |
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
```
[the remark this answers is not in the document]
uriii3: TBD stands for 'To Be Determined', which I think is often used when it is also to be planned. I think it is a nice abbreviation but definitely your call what goes there.

doc/end-of-life.rst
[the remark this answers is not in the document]
uriii3: sure thing

doc/end-of-life.rst
gkoenig-mercator: It could be really useful, I am just unsure if we let it or have to erase it after a certain period.

doc/end-of-life.rst
uriii3: I went for @gkoenig-mercator suggestion! Also, I thought that maybe we could add a `2.1.z` if you want to, but then it seems like it is for all patches... however you prefer!

doc/end-of-life.rst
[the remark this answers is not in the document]
gkoenig-mercator: Well, we have not planned it yet. So I think "Not Planned Yet" without acronym would be more clear and closer to the actual situation. Can you put that please?

doc/end-of-life.rst
```diff
@@ -0,0 +1,79 @@
+=======================================================================
+Versioning and End-of-Life Policy
+=======================================================================
+
+Important information
+-------------------------
+
+The current release, ``v2.3``, no longer supports python 3.9.
```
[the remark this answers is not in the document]
renaudjester: Yes, we can put what you said, Guillaume, in the meantime, and then we adjust when we have a clear decision about this

doc/end-of-life.rst
```diff
@@ -0,0 +1,79 @@
...
+- After EOL, no further fixes, updates, or security patches will be released.
+
+``v1`` is **no longer supported** and should be replaced with ``v2``.
+
+----
+
+Release and End-of-Life Timeline
+---------------------------------
+
++---------------------+--------------+----------------+-------------------+------------------------------------------------------+
```
renaudjester, line 46: check the table in the only documentation: it does not appear in the html :D

doc/end-of-life.rst
```diff
@@ -0,0 +1,79 @@
...
+
+----
+
+Support Policy
+--------------
+
+- Only the **latest minor/patch** in a major branch is supported.
+- After EOL, no further fixes, updates, or security patches will be released.
+
+``v1`` is **no longer supported** and should be replaced with ``v2``.
```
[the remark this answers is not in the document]
gkoenig-mercator: It would be very nicely placed in the "Important Information" section.

doc/end-of-life.rst
```diff
@@ -0,0 +1,79 @@
...
+- **PATCH** – Backward-compatible bug fixes
+
+Example: ``2.1.4`` → major=2, minor=1, patch=4.
+
+----
+
+Support Policy
+--------------
+
+- Only the **latest minor/patch** in a major branch is supported.
```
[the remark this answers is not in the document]
gkoenig-mercator: Well, we used to do it for the v1 but you're right, now we have only one version. So it would be:

"We support only the most recent minor version."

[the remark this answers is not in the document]
uriii3: solved on conversation!

doc/end-of-life.rst
uriii3: corrected

PR #403, opened 2025-08-26
Check order tests matter

uriii3:
cherry pick last commit to check if the order of the tests matter or if it is another commit that is solving the get issues.

----
📚 Documentation preview 📚: https://copernicusmarine--403.org.readthedocs.build/en/403/


uriii3:
This branch was useful to understand the calls of the tests before nd after some refactoring and that was all, closing it.

PR #412, opened 2025-09-22
fix: requests outside dataset interval should raise

renaudjester:
- The default value for `longitude_overlap` should be `False` otherwise we will never raise when the coordinates are out of bounds
- Changed the test `tests/test_warnings_subset_bounds.py::TestWarningsSubsetBounds::test_error_coordinates_out_of_dataset_bounds` because it was not properly testing it, as it was passing and downloading the data.
- Refactor some tests to use `with pytest.raises` instead of try except

### DIFF
--- a/tests/test_describe.py
+++ b/tests/test_describe.py
@@ -3,6 +3,8 @@ import re
 from json import loads
 from unittest import mock
 
+import pytest
+
 from copernicusmarine import (
     CopernicusMarineCatalogue,
     CopernicusMarineServiceNames,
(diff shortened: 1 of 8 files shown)


tests/test_originalGrid_datasets.py
```diff
@@ -210,13 +211,12 @@ def test_originalGrid_error_with_open_dataset(self):
                 coordinates_selection_method="outside",
             )
             assert 1 == 0
```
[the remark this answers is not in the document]
uriii3: why did I write this?

uriii3 (approved):
nice!

tests/test_describe.py
```diff
@@ -93,14 +95,12 @@ def test_describe_with_raise_on_error_error_in_stac(
         self, mocked_requests, caplog
     ):
         with caplog.at_level(logging.DEBUG, logger="copernicusmarine"):
-            try:
+            with pytest.raises(Exception):
```
Theophile-Varnier, line 98: Can't we specify the type of the expected exception here?

tests/test_subset.py
```diff
@@ -228,31 +228,28 @@ def test_lonlat_attributes_when_not_in_arco(self, tmp_path):
                 )
 
     def test_file_format_option(self):
-        response = subset(
+        self.response = subset(
```
Theophile-Varnier, line 231: why?

renaudjester: No reason actually, find it prettier

Issue #441, opened 2025-11-10
copernicusmarine.get returns nothing for static dataset


olehy:
The call `copernicusmarine.get(dataset_id='cmems_mod_glo_phy_anfc_0.083deg_static',overwrite=True,output_directory='.',disable_progress_bar=True,no_directories=True, dry_run=True)`
does return a ResponseGet item with an empty file list using the most recent version 2.2.4 of the toolbox.
```
>>> copernicusmarine.get(dataset_id='cmems_mod_glo_phy_anfc_0.083deg_static',overwrite=True,output_directory='.',disable_progress_bar=True,no_directories=True, dry_run=True)
INFO - 2025-11-10T14:04:01Z - Selected dataset version: "202211"
INFO 2025-11-10 14:04:01,945 services_utils 15584 31012 Selected dataset version: "202211"
INFO - 2025-11-10T14:04:01Z - Selected dataset part: "bathy"
INFO 2025-11-10 14:04:01,946 services_utils 15584 31012 Selected dataset part: "bathy"
INFO - 2025-11-10T14:04:02Z - Listing files on remote server...
INFO 2025-11-10 14:04:02,220 download_original_files 15584 31012 Listing files on remote server...
INFO - 2025-11-10T14:04:03Z - No data to download
INFO 2025-11-10 14:04:03,161 download_original_files 15584 31012 No data to download
ResponseGet(files=[], files_deleted=None, files_not_found=None, number_of_files_to_download=0, total_size=0.0, status='003', message='No data to download from the remote server corresponding to your request.')
```

Using dataset cmems_mod_glo_phy_anfc_0.083deg_PT1H-m delivers many files as expected.

The issue was introduced with version 2.2.4, using 2.2.3 still returns a file.

PR #442, opened 2025-11-10
Check if bucket has folder structure

uriii3:
Although buckets do not have 'folders' we want to make sure that we can list with the desired prefix. Adding the '/' without checking if it was indeed, or not, a resembling folder structure wise architecture would crash the listing.

Hopefully now (it checks that indeed the 'folder exists') the problem will be solved.

Also will fix https://github.com/mercator-ocean/copernicus-marine-toolbox/issues/441 .

### Pull Request Checklist

- [x] Requested code reviews
- [x] Updated the changelog

----
📚 Documentation preview 📚: https://copernicusmarine--442.org.readthedocs.build/en/442/

### DIFF
--- a/doc/changelog/v225.rst
+++ b/doc/changelog/v225.rst
@@ -1,0 +1,12 @@
+Release 2.2.5
+==============
+
+Patch that fixes some issues when accessing static and OMI datasets.
+
+Get
+------
+
+Fixes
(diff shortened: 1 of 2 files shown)


renaudjester (commented):
Missing changelog otherwise looks good to me

PR #446, opened 2025-11-18
feat: split-on as a subcommand

renaudjester:
fix [CMT-116](https://cms-change.atlassian.net/browse/CMT-116) after redefining the needs

It has many changes, as this new subcommand asked for some refactoring: 

Main features: 

- `split-on` as a subcommand: can be used like this 
```copernicusmarine subset [SUBSET-OPTIONS] split-on [SPLIT-ON-OPTIONS]```
- `subset_split_on` as a Python function, that accepts all the arguments of `subset` in the `kwargs`
- `open_dataset` and `read_dataframe` should now work with `staging`
- `split-on` only uses the metadata to create the different keys and then starts a subset for each split (without the metadata parsing step)
- specific page for the split-on in the doc

Refactor:

- Merged `split-on` tests to follow another recent refactor of the tests (not distinguishing command line and python interface, but more by command_
- Now, all `subset` commands and Python functions create a `SubsetRequest`, which allows for centralising some checks and some modifications of the parameters (like the login)
- Hence, all `subset` core functions take as input this `SubsetRequest` object. 
- Deleted the `load_utils` files and the `LoadRequest` object to standardise inputs 
- Added an `open_dataset_function` and `read_dataframe_function` for clarity and grouped different logics to avoid code repetition and centralise changes. 

### Pull Request Checklist

- [x] Requested code reviews
- [x] Updated the changelog

[CMT-116]: https://cms-change.atlassian.net/browse/CMT-116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
📚 Documentation preview 📚: https://copernicusmarine--446.org.readthedocs.build/en/446/

### DIFF
--- a/doc/changelog/v230.rst
+++ b/doc/changelog/v230.rst
@@ -29,7 +29,7 @@ New Features
 * Inverting the order of the arguments for minimum and maximum values in ``subset`` will now raise an error. Except for the longitude, it raises a warning (see item below). This helps users avoid mistakes and ensures more consistent behavior.
 * Requests across the antemeridian line (180° longitude) can now be made by specifying a minimum value greater than the maximum value (except for sparse datasets). The Toolbox will automatically shift the window to the left or right of the antemeridian line, depending on the requested area. See the :ref:`Subset documentation <about-longitude-range>` for details.
 * Subset results now preserve more attributes, as the Toolbox passes through attributes from the server data. Previously, only a whitelisted set of attributes was included.
-* Added a new command ``split-on`` to the `subset` command to split the output files based on time or variables when subsetting. See :ref:`subset usage page <subset-split-on>` for more details.
+* Added a new subcommand ``split-on`` to the ``subset`` command to split the output files based on time or variables when subsetting. See :ref:`subset usage page <subset-split-on>` for more details.
 * The Toolbox now creates a temporary file when subsetting data to avoid keeping incomplete files if an error occurs during the subsetting process.
 
 Fixes
--- a/doc/usage/subset-split-on-usage.rst
+++ b/doc/usage/subset-split-on-usage.rst
@@ -62,7 +62,6 @@ The Toolbox uses Python's ``concurrent.futures.ProcessPoolExecutor`` to manage p
 
   response = copernicusmarine.subset_split_on(
       dataset_id="cmems_mod_glo_phy-all_my_0.25deg_P1D-m",
-      dataset_id=dataset_id,
       start_datetime="2021-01-01",
       end_datetime="2025-01-03",
       on_time="year",
--- a/tests/test_subset_split_on.py
+++ b/tests/test_subset_split_on.py
@@ -30,7 +30,6 @@ class TestSubsetSplitOn:
             output_directory=tmp_path,
             disable_progress_bar=True,
         )
-        assert isinstance(res, list)
         assert len(res) == 2
         ds_2022_path = res[0].file_path
         assert os.path.exists(ds_2022_path)
@@ -103,7 +102,6 @@ class TestSubsetSplitOn:
             output_directory=tmp_path,
             dry_run=True,
         )
-        assert isinstance(res, list)
         assert len(res) == 4
         filenames = [f.filename for f in res]
         for period in [
@@ -129,7 +127,6 @@ class TestSubsetSplitOn:
             on_variables=True,
             output_directory=tmp_path,
         )
-        assert isinstance(res, list)
         assert len(res) == 3
         assert os.path.exists(
             os.path.join(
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: copernicusmarine/command_line_interface/command_subset_split_on.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+)
+@click.pass_context
+@log_exception_and_exit
+def split_on(
+    ctx,
+    on_variables: bool,
+    on_time: Optional[SplitOnTimeOption],
+    concurrent_processes: Optional[int],
+):
+    """Function description"""
```
uriii3: not needed I guess

copernicusmarine/command_line_interface/command_subset_split_on.py
```diff
@@ -0,0 +1,113 @@
...
+@click.option(
+    "--concurrent-processes",
+    type=click.IntRange(1, None),
+    default=None,
+    help=SUBSET_SPLIT_ON["CONCURRENT_PROCESSES_HELP"],
+)
+@click.pass_context
+@log_exception_and_exit
+def split_on(
+    ctx,
```
uriii3: maybe write the whole word `context` as guidance? I dunno

copernicusmarine/core_functions/request_structure.py
```diff
@@ -68,40 +76,58 @@ class SubsetRequest(BaseModel):
...
+    chunk_size_limit: int = -1
 
-    def update(self, new_dict: dict):
+    def update(self, new_dict: dict) -> "SubsetRequest":
         filtered_dict = {
-            key: value for key, value in new_dict.items() if value is not None
+            key: value
+            for key, value in new_dict.items()
+            if value is not None
+            and not (isinstance(value, (list, tuple, str)) and not value)
```
uriii3: what does it imply the last `and not value`? that if it is True, it will not update anything? it looks like this `if` is filtering a lot of the options we would want to update, no? I don't understand.

copernicusmarine/core_functions/subset.py
```diff
@@ -167,87 +74,72 @@ def subset_function(
...
+        )
+        if subset_request.coordinates_selection_method not in [
+            "inside",
+            "strict-inside",
+        ]:
+            logger.warning(
+                f"coordinates-selection-method "
+                f"{subset_request.coordinates_selection_method} "
+                "is not supported for sparse data. "
+                "Using 'inside' by default."
```
uriii3: we don't need to change the value, it will be correctly understood inside the parse download function, right? just throw a warning and that is good enough.

copernicusmarine/python_interface/subset_split_on.py
```diff
@@ -0,0 +1,86 @@
...
+@log_exception_and_exit
+def subset_split_on(
+    on_variables: bool = False,
+    on_time: Optional[SplitOnTimeOption] = None,
+    concurrent_processes: Optional[int] = None,
+    **kwargs,
+) -> list[ResponseSubset]:
+    """
+    Extract a subset of data from a specified dataset using given parameters
+    and split the output files based on variable names, time intervals or both.
```
uriii3: I don't know how to write (or if we need to) the fact that you need at least one of them. I think the error is quite explanatory, so maybe it is enough to just leave it as it is right now.

copernicusmarine/core_functions/request_structure.py
uriii3: Wow, okey, a big complication to be able to include values=0, but it makes much more sense now. It is not easy to grab the logic though, or at least not for me hahaha.

copernicusmarine/python_interface/open_dataset.py
```diff
@@ -127,75 +118,47 @@ def open_dataset(
     xarray.Dataset
         The loaded xarray dataset.
     """  # noqa
-    start_datetime = homogenize_datetime(start_datetime)
-    end_datetime = homogenize_datetime(end_datetime)
+    if variables is not None:
+        _check_type(variables, list, "variables")
```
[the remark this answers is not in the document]
uriii3: aaaah, okey... so we want to hint the users about the correct way of passing the variables, right? pretty nice idea! I saw that we were already using it but with everything else in between I didn't even notice!

I think this makes it even more clear what the tidiness of the code you wrote does! Amazing to see it!

uriii3 (approved):
For me, correct and add the tests we spoke about and it is approved! the general idea and structure seems fine for me and I like the rework on all the parts you touched. If there is any specific thing that needs to be improved I think we can manage with the beta version and using the toolbox ourselves before the official release.

Again, congrats on the rework, it is amazing to see how you managed everything! 🚀

PR #459, opened 2025-12-28
Allow passing `max_concurrent_requests=0` to avoid `ThreadPoolExecutor`

juntyr:
@dbazin-eo and I are currently trying to get copernicusmarine to get working inside https://github.com/climet-eu/lab in https://github.com/climet-eu/lab/issues/18.

The Online Laboratory for Climate Science and Meteorology (https://github.com/climet-eu/lab) is based on Pyodide which runs Python inside WebAssembly, where the ThreadPoolExecutor is not available. This PR allows users to set max_concurrent_requests=0 to explicitly avoid the ThreadPoolExecutor.

fix [CMT-364](https://cms-change.atlassian.net/browse/CMT-364)

----
📚 Documentation preview 📚: https://copernicusmarine--459.org.readthedocs.build/en/459/

[CMT-364]: https://cms-change.atlassian.net/browse/CMT-364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/copernicusmarine/core_functions/sessions.py
+++ b/copernicusmarine/core_functions/sessions.py
@@ -105,11 +105,16 @@ class ConfiguredBoto3Session:
         )
         self.use_threads = COPERNICUSMARINE_USE_THREADS
 
+    def close(self):
+        self.s3_client.close()
+        if self.s3_resource:
+            self.s3_resource.meta.client.close()
+
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        pass
+        self.close()
 
     def download_file(
         self, bucket_name: str, object_key: str, file_path: str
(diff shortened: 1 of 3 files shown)


renaudjester:
Hi @juntyr! 

Thanks for the PR. The idea looks good to me, and the code too. We will do a review and merge after we release v2.3.0 of the Toolbox, probably around next week. 

Have a great day!

renaudjester:
Hello again @juntyr,

I discussed this with @dbazin-eo, and it seems that there could also be a problem with the fact that, by default, boto3 uses threads. 

So I propose that: 
- We add an environment variable that will disable any threading/multiprocessing computation
- We will add documentation
- We can probably keep the `max_concurrent_request=0` to be consistent with the `get` command

If it suits you, I can do all of this on this branch! Otherwise, I can open another PR.

juntyr:
That sounds good to me! Btw what browser are you using?

renaudjester:
> That sounds good to me! Btw what browser are you using?

Google Chrome!

I cannot seem to be able to open the downloaded data, though 🤔

PR #470, opened 2026-01-21
Add h5py as an explicit dependency

uriii3:
Some of the binaries actions were failing due to a recent release on `h5netcdf`, which removed the installation of `h5py` when installing the library. 

You can see the issue explained here: https://github.com/h5netcdf/h5netcdf/issues/310

From that, we need to also install `h5py` as a dependency (maybe optional?) for the toolbox.

This would solve the binaries actions.

### Pull Request Checklist

- [x] Requested code reviews
- [ ] Updated the changelog

----
📚 Documentation preview 📚: https://copernicusmarine--470.org.readthedocs.build/en/470/

### DIFF
--- a/tests/__snapshots__/test_dependencies_updates.ambr
+++ b/tests/__snapshots__/test_dependencies_updates.ambr
@@ -5,7 +5,12 @@
     'boto3': '>=1.26',
     'click': '>=8.0.4,!=8.2.0,<8.3.0',
     'dask': '>=2022',
-    'h5netcdf': '^1.4.0',
+    'h5netcdf': DynamicInlineTableDict({
+      'extras': list([
+        'h5py',
+      ]),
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -25,6 +25,7 @@ numpy = ">=1.23.0"
 pydantic = "^2.9.1"
 h5netcdf = "^1.4.0"
 arcosparse = "^0.4.2"
+h5py = "^3.15.1"
```
renaudjester: Why don't you install it through h5netcdf instead? As described in the issue

uriii3: I wouldn't know how to do it when declaring the dependencies here in the pyproject!

uriii3: maybe it is trickier than I thought! let's talk it today

PR #495, opened 2026-03-24
fix: update attributes netcdf sparse subset

renaudjester:
Update the netcdf attributes

### Pull Request Checklist

- [x] Requested code reviews
- [ ] Updated the changelog

----
📚 Documentation preview 📚: https://copernicusmarine--495.org.readthedocs.build/en/495/


copernicusmarine/download_functions/download_sparse.py
```diff
@@ -550,7 +546,7 @@ def _platform_dataframe_to_dataset(
         aggfunc="first",
     )
     pivot.columns = [
-        f"{var}_qc" if metric == "value_qc" else var
+        f"{var}_QC" if metric == "value_qc" else var
         for metric, var in pivot.columns
     ]
```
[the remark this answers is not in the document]
uriii3: update the documentation if it says so, right?

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -681,69 +700,73 @@ def _add_attributes_to_dataset(
...
+        ds[vertical_axis].attrs["axis"] = "Z"
 
     # pressure attributes
-    ds["pressure"].attrs["units"] = "dbar"
-    # ds["pressure"].attrs["axis"] = "Z"
-    ds["pressure"].attrs["positive"] = "down"
-    ds["pressure"].attrs["standard_name"] = "reference_pressure"
+    if "pressure" in ds.coords:
+        ds["pressure"].attrs["units"] = "dbar"
+        # ds["pressure"].attrs["axis"] = "Z"
```
uriii3: remove

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -588,6 +603,8 @@ def _build_netcdf_encoding(
             "complevel": compression_level,
             "contiguous": False,
             "shuffle": True,
+            "scale_factor": 1.0,
+            "add_offset": 0.0,
```
uriii3, line 607: hardcoded like this? why?

copernicusmarine/download_functions/download_sparse.py
uriii3: remove?

uriii3 (commented):
can you remind me the 'depth from producer' and the relation between when it is depth(elevation) or pressure?

copernicusmarine/download_functions/download_sparse.py
renaudjester: I think we can optimise it later :D

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -527,19 +519,23 @@ def _dataframe_to_netcdf_per_platform(
 
 def _platform_dataframe_to_dataset(
     platform_df: pd.DataFrame,
-    index_cols: list[str],
-    aux_cols: list[str],
     metadata_cols: list[str],
+    vertical_axis: VerticalAxis,
```
uriii3, line 523: this we set to 'depth' by default, but does not necessarily mean that we have the dimension, no? do we always have depth OR elevation? or we could have pressure as the 'Z' axis? then this works?

uriii3 (commented):
As I understand from the code and ticket, we have:
- depth or elevation
- maybe we have pressure -> which should be axis Z if none of the above (?)
What has happened to `depth_from_producer` ?

Theophile-Varnier (approved):

copernicusmarine/download_functions/download_sparse.py
renaudjester: We always have the dimension :D the ARCO data will always have depth even if the producer doesn't provide it

uriii3 (approved):

PR #498, opened 2026-04-14
feat: take into consideration sparse type

renaudjester:
fix [CMT-228](https://cms-change.atlassian.net/browse/CMT-228)

Using the attribute "sparse_type" in the metadata to determine if we should have a netcdf with dimension (time, depth_level) or just (time) 

### Pull Request Checklist

- [x] Requested code reviews
- [x] Updated the changelog

[CMT-228]: https://cms-change.atlassian.net/browse/CMT-228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/copernicusmarine/download_functions/download_sparse.py
+++ b/copernicusmarine/download_functions/download_sparse.py
@@ -478,6 +478,7 @@ def _dataframe_to_netcdf_per_platform(
                 platform_df,
                 metadata_cols=metadata_cols,
                 vertical_axis=vertical_axis,
+                arco_sparse_type=service.arco_sparse_type,
             )
 
             ds = _add_attributes_to_dataset(
@@ -521,6 +522,7 @@ def _platform_dataframe_to_dataset(
     platform_df: pd.DataFrame,
     metadata_cols: list[str],
     vertical_axis: VerticalAxis,
+    arco_sparse_type: str | None,
 ) -> xr.Dataset:
     """
     ``depth_level`` is an integer index: for each time step, depth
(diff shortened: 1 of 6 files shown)


copernicusmarine/download_functions/download_sparse.py
```diff
@@ -537,28 +539,37 @@ def _platform_dataframe_to_dataset(
...
+    spatial_and_aux = obs_df.groupby(pivot_columns)[aux_cols].first()
 
     merged = pd.concat([pivot, spatial_and_aux], axis=1)
     merged = merged.reset_index()
-    merged["depth_level"] = merged.groupby("time").cumcount()
     merged["time"] = merged["time"].apply(
         lambda x: datetime_to_timestamp(datetime_parser(x), unit="s")
     )
-    merged = merged.set_index(["time", "depth_level"])
+    if arco_sparse_type != "cmemsAltimetry":
```
Theophile-Varnier, line 568: Maybe take this hard coded value somewhere in an ENUM or a const variable defined in a helper file?

Theophile-Varnier (approved):

copernicusmarine/download_functions/download_sparse.py
uriii3, line 568: I would say that it is also weird that you invert the logic here and up (the other if, checks if the first is equal to `cmemsAltimetry` and then does the other case. Inverting the 2 makes it a bit harder to assmilate the logic.

copernicusmarine/download_functions/download_sparse.py
```diff
@@ -620,6 +636,11 @@ def _add_attributes_to_dataset(
     product_id: str | None,
     service: CopernicusMarineService,
 ) -> xr.Dataset:
+    def unpack(value_from_df):
```
uriii3, line 639: and this?

uriii3 (commented):
Were we already reading the `service.arco_sparse_type` variable? I have not seen exactly where we add the reading of such variable!

uriii3 (approved):

copernicusmarine/download_functions/download_sparse.py
renaudjester: It's defined in the description of the `arco_sparse_type` field for the CopernicusMarineService! I don't want to do a literal because pydantic would raise if it wasn't the exact literal which I cannot really control since the value comes from the metadata

I don't know if I am clear :D

renaudjester:
> Were we already reading the `service.arco_sparse_type` variable? I have not seen exactly where we add the reading of such variable!

In the catalogue_parser/models.py

copernicusmarine/download_functions/download_sparse.py
renaudjester: Basically, in xarray:

If you have two dimensions (for example, `(time, depth_level)`) and you retrieve a value from a coordinate (like `latitude`), you will get a list: `np.array([value, ...])`.

If you have one dimension (for example, `(time)`) and you retrieve a value from a coordinate (like `latitude`) you will get a number: `np.array(value)`, so I am treating those two cases

PR #507, opened 2026-05-26
fix: dataset can be describe if it is in several products

renaudjester:
Make sure that if a dataset is in two products, then it works anyway. 

Let's say datasetID is in productID1 and productID2, then the dataset_product_mapping.json we would have:
```
datasetID: "productID1,productID2"
```
It works with `get` and `subset` but would fail when asking for this specific dataset with `describe`: 

```
# error
copernicusmarine describe --dataset-id datasetID
```

fix [CMT-398](https://cms-change.atlassian.net/browse/CMT-398)

### Pull Request Checklist

- [x] Requested code reviews
- [x] Updated the changelog

----
📚 Documentation preview 📚: https://copernicusmarine--507.org.readthedocs.build/en/507/

[CMT-398]: https://cms-change.atlassian.net/browse/CMT-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -453,24 +457,25 @@ def parse_catalogue(
...
-                and product_id_from_mapping != force_product_id
-            ):
+            product_ids = product_id_from_mapping.split(",")
+            if force_product_id and force_product_id not in product_ids:
                 raise DatasetIsNotPartOfTheProduct(
                     force_dataset_id, force_product_id
                 )
-            force_product_id = product_id_from_mapping
+            elif force_product_id:
+                product_ids = [force_product_id]
```
uriii3, line 474: what is the logic for a dataset to be in different `product_ids`? I understand that, for now, if we force a product id is always only one, but depending on the case of why that dataset is split between multiple products, i understand that it would not be super weird to specify maybe multiple products?(i know we can only just ask for the dataset id and not force the product, which can be the simple answer to that).

tests/resources/mock_stac_catalog_WAW3/mock_dataset_DUPLICATE_nwshelf_P1D_m_202012.py
uriii3, line 1: is there any difference here between this and the one from before apart from the product id?

tests/resources/mock_stac_catalog_WAW3/mock_product_DUPLICATE_nwshelf.py
uriii3, line 1: could you not point the dataset to an already existing product? is there a reason to specifically also build another product?

uriii3 (approved):

copernicusmarine/catalogue_parser/catalogue_parser.py
```diff
@@ -411,7 +413,7 @@ def fetch_product_items(
 
 def fetch_all_products_items(
     connection: JsonParserConnection,
-    force_product_id: str | None,
+    force_product_ids: list[str] | str | None,
```
Theophile-Varnier: why do we still allow `str` here and not in `fetch_product_items`?

tests/resources/mock_stac_catalog_WAW3/mock_product_DUPLICATE_nwshelf.py
renaudjester: Yeah I could but I thought this would be clearer :D

tests/resources/mock_stac_catalog_WAW3/mock_dataset_DUPLICATE_nwshelf_P1D_m_202012.py
renaudjester: Also the links I think!

copernicusmarine/catalogue_parser/catalogue_parser.py
renaudjester: It makes the variable a bit easier to use! in `fetch_all_products_items` we just pass the variable but in `fetch_product_items` we actually use it and do some tests with it so I thought it would be more practical to have only one typing, to clarify the function

Theophile-Varnier (approved):

PR #530, opened 2026-09-22
fix: binaries tests are failing

renaudjester:
Binaries tests seem to be failing in the last commit to main

### Pull Request Checklist

- [ ] Requested code reviews
- [ ] Updated the changelog


uriii3 (approved):
Yep... actually i was thinking that maybe we could end up writting a conf file for them, as they all need many times this 'copy-metadata' and others, but maybe if i end up solving one of those in the future i can do it! Good to know you are also good doing this now 👀

